PR #21, opened 2023-08-03
EVEREST-165: monitoring config

michal-kralik:
EVEREST-165


api/v1alpha1/monitoringconfig_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// CredentialsSecretName is the name of the secret with credentials.
+	CredentialsSecretName string `json:"credentialsSecretName,omitempty"`
+	// PMM is configuration for the PMM monitoring type.
+	PMM PMMConfig `json:"pmm,omitempty"`
+}
+
+// PMMConfig is configuration of the PMM monitoring type.
+type PMMConfig struct {
+	// URL is url to the monitoring instance.
+	URL string `json:"url,omitempty"`
```
oksana-grishchenko, line 44: url is also generated as optional in the openapi spec because of that `omitempty`. I believe it should be required.

e2e-tests/tests/monitoringconfig_pxc/01-assert.yaml
```diff
@@ -0,0 +1,27 @@
...
+data:
+  pmmserverkey: YWJjZGVm
+---
+apiVersion: pxc.percona.com/v1
+kind: PerconaXtraDBCluster
+metadata:
+  name: pxc-mc
+spec:
+  secrets:
+    users: everest-secrets-pxc-mc
```
oksana-grishchenko, line 27: The `MonitoringConfig` CR is missing in the e2e tests asserts

oksana-grishchenko (commented):
Great Job! 

Running `make bundle` and make `deploy/bundle.yaml` showed that there are some uncommitted generated changes, could you please check it?

config/samples/everest_v1alpha1_monitoringconfig.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  labels:
+    app.kubernetes.io/name: monitoringconfig
+    app.kubernetes.io/instance: monitoringconfig-sample
+    app.kubernetes.io/part-of: everest-operator
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: everest-operator
+  name: monitoringconfig-sample
+spec:
+  # TODO(user): Add fields here
```
recharte, line 12: I know I'm also guilty of completely ignoring this in my previous PRs but we need to start filling in these samples.

api/v1alpha1/monitoringconfig_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// Image is a Docker image name to use for deploying PMM client. Defaults to using the latest version.
+	Image string `json:"image,omitempty"`
```
recharte, line 46: I actually think we might be able to go without this field.
I think we should query the PMM server for the version it’s running and use the appropriate image.
We can offload this to the user and force it to specify it (it would be great not having to though) but I don't think that using the `latest` tag (or any default tag) is a good idea.

recharte (commented):
This is not a full review yet, I have only looked at the API but I wanted to share some feedback as soon as possible.

api/v1alpha1/databasecluster_types.go
```diff
@@ -218,8 +201,8 @@ type DatabaseClusterSpec struct {
 	DataSource *DataSource `json:"dataSource,omitempty"`
 	// Backup is the backup specification
 	Backup Backup `json:"backup,omitempty"`
-	// Monitoring is the monitoring specification
-	Monitoring Monitoring `json:"monitoring,omitempty"`
+	// MonitoringConfigName is the name of a monitoringConfig CR.
+	MonitoringConfigName string `json:"monitoringConfigName,omitempty"`
```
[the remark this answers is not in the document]
recharte: I don't see this as a big deal but I'd go with an `Enabled` field.

api/v1alpha1/databasecluster_types.go
```diff
@@ -187,22 +186,6 @@ type PMMSpec struct {
 	Password      string `json:"password,omitempty"`
 }
 
-// Monitoring contains monitoring settings.
-type Monitoring struct {
-	// Enabled is a flag to enable monitoring
-	Enabled   bool                        `json:"enabled"`
-	PMM       *PMMSpec                    `json:"pmm,omitempty"`
-	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
```
recharte, line 195: Was the removal of the resources field a conscious decision? Wouldn't it be useful to be able to set resource limits for the monitoring container?

controllers/databasecluster_controller.go
```diff
@@ -715,6 +746,56 @@ func (r *DatabaseClusterReconciler) reconcilePSMDB(ctx context.Context, req ctrl
...
+	ctx context.Context, database *everestv1alpha1.DatabaseCluster,
+	monitoring *everestv1alpha1.MonitoringConfig, secretKey string,
+) (string, error) {
+	var secret *corev1.Secret
+	secretData := ""
+
+	if monitoring.Spec.CredentialsSecretName != "" {
+		secret = &corev1.Secret{}
+		err := r.Get(ctx, types.NamespacedName{
+			Name:      monitoring.Spec.CredentialsSecretName,
```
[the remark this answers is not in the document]
michal-kralik: My goal was to abstract which field to use for the secret name.
Do you think it's better to just pass the secret name field when calling this function?

api/v1alpha1/monitoringconfig_types.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// MonitoringConfigStatus defines the observed state of MonitoringConfig.
+type MonitoringConfigStatus struct{}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
```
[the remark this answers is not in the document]
michal-kralik: That's a great suggestion! I wasn't aware of this feature.

api/v1alpha1/databasecluster_types.go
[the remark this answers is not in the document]
michal-kralik: Removing the `monitoringConfigName` shall do the trick. I think it's better than introducing a new `Enabled` field.

api/v1alpha1/databasecluster_types.go
michal-kralik: No, this is a good catch.
I did not see it in the CR samples and thought it's not available, but it is. Added back.

api/v1alpha1/monitoringconfig_types.go
michal-kralik: I agree it would be best to leave it out.
I don't think it's possible at the moment though, but shall be possible in the future.

At the moment we need to give the user the option to configure the version of `pmm-client` since we don't have the version detection.
The detection mechanism needs to regularly check what the version is and change the image name accordingly. I think it's a bigger topic and we may need to include the PMM team in the discussion to figure out how to approach this.

api/v1alpha1/monitoringconfig_types.go
```diff
@@ -0,0 +1,79 @@
...
+
+// MonitoringConfigSpec defines the desired state of MonitoringConfig.
+type MonitoringConfigSpec struct {
+	// Type is type of monitoring.
+	// +kubebuilder:validation:Enum=pmm
+	Type MonitoringType `json:"type"`
+	// CredentialsSecretName is the name of the secret with credentials.
+	CredentialsSecretName string `json:"credentialsSecretName"`
+	// Resources defines resource limitations for the monitoring.
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
```
recharte, line 39: I don't currently have an answer for this but I wanted to start the discussion.
Should the `Resources` be part of the `MonitoringConfig` or should it be part of the `DatabaseCluster` that references it?
Phrasing this in a different way, would we ever need to set different resource limits for the same Monitoring instance? Would a PMM client container need more resources in DatabaseCluster A than in DatabaseCluster B?

api/v1alpha1/monitoringconfig_types.go
recharte: I agree with you but in this case I think we should then make this field mandatory (remove the omitempty) because we can't guarantee that the `latest` client tag is compatible with the server version the user is running.

e2e-tests/tests/monitoringconfig_pxc/01-assert.yaml
recharte: Yes I think you're right Michal, I actually thought about this before because we have this across all other tests.
Currently, there isn't much point in testing what you applied.
However, if we end up adding logic to change the spec and add for example the default values we might have a use case for also testing the resource you just applied. This is just far fetched right now.

api/v1alpha1/monitoringconfig_types.go
michal-kralik: I think you're right. This shall sit with the `DatabaseCluster` and not `MonitoringConfig`.
Different DBs may require more/less resources.

api/v1alpha1/monitoringconfig_types.go
michal-kralik: Updated

api/v1alpha1/databasecluster_types.go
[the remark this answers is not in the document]
michal-kralik: That's correct. I updated the implementation a couple minutes ago ;)

api/v1alpha1/monitoringconfig_types.go
```diff
@@ -0,0 +1,79 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// MonitoringType is a type of monitoring.
+type MonitoringType string
+
+const (
+	// PMM represents monitoring via PMM.
+	PMM MonitoringType = "pmm"
```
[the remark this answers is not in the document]
michal-kralik: I usually prefer `PMM` instead of `PMMMonitoringType` as I feel it's too verbose and `PMM` has a type of `MonitoringType`.
But I'm happy to change it. It's only a matter of personal preference.

michal-kralik:
> Instead of using pmm-client:latest can we call `v1/version` endpoint on the PMM and fill image field? WDYT?

@recharte suggested we don't provide any default since the user needs to make a decision which version to use.
The user shall change the version manually on upgrades as well.

I believe at some point we may need to create a mechanism to automatically detect and change the version if needed. Not sure if it's in the scope at this point though.

gen1us2k:
The response from http://127.0.0.1/v1/version endpoint (it requires authentication and we have credentials) returns the following

```
{
  "version": "2.38.0",
  "server": {
    "version": "2.38.0",
    "full_version": "2.38.0-20.2306291648.03d5997.el9",
    "timestamp": "2023-06-29T16:48:27Z"
  },
  "managed": {
    "version": "2.38.0",
    "full_version": "03d5997be9810a6c8503159b09a64153b77010bc",
    "timestamp": "2023-06-29T16:48:30Z"
  },
  "distribution_method": "DOCKER"
}
```

We can get the version and fill the image easily because we have the endpoint, credentials, and everything else of the connected PMM instance. I see no reason not to implement it because it will improve the UX.

Another thing is that it misses the generation of a new token for a created db cluster.

gen1us2k (changes requested):
The response from http://127.0.0.1/v1/version endpoint (it requires authentication and we have credentials) returns the following

```
{
  "version": "2.38.0",
  "server": {
    "version": "2.38.0",
    "full_version": "2.38.0-20.2306291648.03d5997.el9",
    "timestamp": "2023-06-29T16:48:27Z"
  },
  "managed": {
    "version": "2.38.0",
    "full_version": "03d5997be9810a6c8503159b09a64153b77010bc",
    "timestamp": "2023-06-29T16:48:30Z"
  },
  "distribution_method": "DOCKER"
}
```

We can get the version and fill the image easily because we have the endpoint, credentials, and everything else of the connected PMM instance. I see no reason not to implement it because it will improve the UX.

Another thing is that it misses the generation of a new token for a created db cluster.

gen1us2k:
I think we need to generate a new token for every db cluster because it makes everything easier to manage when we changed the secret for the monitoringconfig we need to update every secret for every database cluster it uses.

gen1us2k:
Also tests are failing because of 
```
MonitoringConfig.everest.percona.com "test-mc" is invalid: spec.pmm.image: Required value
```

controllers/databasecluster_controller.go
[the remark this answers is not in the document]
oksana-grishchenko: Ok, that's a good reason to have it as it is

oksana-grishchenko (approved):
Approving since my concerns were resolved. 

Regarding the concerns about token generation and versions configuration - it's a good point, I believe we'll come to it eventually, the question is wether we want it in the scope of this PR.

michal-kralik:
> I think we need to generate a new token for every db cluster because it makes everything easier to manage when we changed the secret for the monitoringconfig we need to update every secret for every database cluster it uses.

I believe it's better to reuse the same token. This way the user can disable a single token and revoke access to monitoring altogether.

gen1us2k (changes requested):

config/rbac/monitoringconfig_editor_role.yaml
```diff
@@ -0,0 +1,31 @@
+# permissions for end users to edit monitoringconfigs.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: clusterrole
+    app.kubernetes.io/instance: monitoringconfig-editor-role
+    app.kubernetes.io/component: rbac
```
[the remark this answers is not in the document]
michal-kralik: I think they're automatically generated by the `operator` cli

controllers/databasecluster_controller.go
```diff
@@ -677,10 +679,34 @@ func (r *DatabaseClusterReconciler) reconcilePSMDB(ctx context.Context, req ctrl
...
 
-		if database.Spec.Monitoring.PMM != nil && database.Spec.Monitoring.PMM.Image != "" {
+		monitoring := &everestv1alpha1.MonitoringConfig{}
+		if database.Spec.Monitoring != nil && database.Spec.Monitoring.MonitoringConfigName != "" {
+			err := r.Get(ctx, types.NamespacedName{
+				Namespace: req.NamespacedName.Namespace,
+				Name:      database.Spec.Monitoring.MonitoringConfigName,
+			}, monitoring)
+			if err != nil {
+				return err
```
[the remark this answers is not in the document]
michal-kralik: It can be disabled by omitting the `monitoring` field or providing an empty `monitoringConfigName`

api/v1alpha1/databasecluster_types.go
```diff
@@ -16,7 +16,7 @@
 package v1alpha1
 
 import (
-	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
gen1us2k, line 19: while working with Kubernetes client-go having this import using `corev1` makes it more clear which API and package is responsible for the structures you use.

config/crd/patches/cainjection_in_monitoringconfigs.yaml
```diff
@@ -0,0 +1,7 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
+  name: monitoringconfigs.everest.percona.com
```
gen1us2k, line 7: Please add a new CRD to the cleanup step in the pipeline https://github.com/percona/everest-operator/blob/main/.github/workflows/ci.yml#L155

config/rbac/monitoringconfig_editor_role.yaml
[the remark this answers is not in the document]
gen1us2k: There are no changes in role.yml. Take a look here 
https://github.com/percona/everest-operator/pull/22/files#diff-6d4c4e29ec096149781536bb9118fde8242c1901df764e3eda3019ac8503d6c7
We need to have it in the database controller

e2e-tests/tests/monitoringconfig_pg/01-assert.yaml
```diff
@@ -0,0 +1,34 @@
...
+spec:
+  pmm:
+    enabled: true
+    serverHost: http://localhost
+    image: percona/pmm-client:latest
+    secret: everest-secrets-pg-mc
+    resources:
+      limits:
+        memory: "128Mi"
+        cpu: "500m"
```
gen1us2k, line 34: It misses asserts for the status field that a database cluster getting ready. Having a ready cluster will improve tests and other steps that ensure that it changes something when we change a secret or monitoring config should also assert that a database cluster changed its state to `initializing`.

Another solution to ensure that reconcile loop was triggered is to check the generation field in the metadata and ensure that it increments. In that case tests will check that it actually changes something in the databasecluster CR

gen1us2k: However, a database cluster may not change the state to `initializing` and generation may be not the best thing to run asserts against.
https://stackoverflow.com/a/66092577 you can choose the best option from this list

gen1us2k (changes requested):

api/v1alpha1/databasecluster_types.go
```diff
@@ -177,30 +177,12 @@ type Backup struct {
...
-	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
-	//nolint:godox,dupword,gocritic
-	// TODO migrate to the MonitoringConfig CR approach.
-	//// MonitoringConfigName is the name of the MonitoringConfig CR that defines
-	//// the monitoring configuration
-	//MonitoringConfigName string `json:"monitoringConfigName,omitempty"`
-	//// Resources are the resource requirements for the monitoring container.
-	//// If not set, resource limits are not imposed
-	//Resources Resources `json:"resources,omitempty"`
+	// ConfigName is the name of a monitoringConfig CR.
```
recharte, line 182:
```suggestion
	// MonitoringConfigName is the name of a monitoringConfig CR.
```

config/samples/everest_v1alpha1_monitoringconfig.yaml
```diff
@@ -0,0 +1,21 @@
...
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: everest-operator
+  name: monitoringconfig-sample
+spec:
+  type: pmm
+  credentialsSecretName: my-secret-name
+  resources:
+    limits:
+      cpu: "300m"
+      memory: "0.5G"
```
recharte, line 17:
```suggestion
```

controllers/databasecluster_controller.go
```diff
@@ -716,6 +742,55 @@ func (r *DatabaseClusterReconciler) reconcilePSMDB(ctx context.Context, req ctrl
...
+			Name:      secretName,
+			Namespace: database.Namespace,
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: data,
+	}
+
+	if secretName == "" {
+		secret.ObjectMeta.GenerateName = generateName
+	}
```
recharte, line 789: This is interesting :thinking:

recharte (commented):
@michal-kralik will the outcomes of the offline discussion still be reflected in this PR?

config/rbac/monitoringconfig_editor_role.yaml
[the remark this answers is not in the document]
michal-kralik: yeah, no idea. Is this file expected to be updated automatically or manually?

controllers/databasecluster_controller.go
```diff
@@ -542,7 +364,7 @@ func (r *DatabaseClusterReconciler) reconcilePSMDB(ctx context.Context, req ctrl
 			return err
 		}
 
-		psmdb.Spec = defaultPSMDBSpec
+		psmdb.Spec = *r.defaultPSMDBSpec()
```
gen1us2k, line 367: Why do we need to return it as a reference and use it as a value?

config/rbac/monitoringconfig_editor_role.yaml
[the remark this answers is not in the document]
gen1us2k: It should be generated automatically. Just add 
```
//+kubebuilder:rbac:groups=everest.percona.com,resources=monitoringconfigs,verbs=get;list;watch;create;update;patch;delete
```
to databasecluster_controller and run regeneration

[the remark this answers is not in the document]
michal-kralik: That fixed it, thanks

controllers/databasecluster_controller.go
michal-kralik: It saves one copy of the struct by value.

gen1us2k (approved):

PR #42, opened 2023-08-18
EVEREST-181 Scheduled backups

oksana-grishchenko:
EVEREST-181  Scheduled backups

`DatabaseClusterBackup` resource is created when an upstream backup resource without an owner reference appears.


controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +160,206 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+			logger.Error(err, fmt.Sprintf("Failed to create DatabaseClusterBackup %s", obj.GetName()))
+		}
+	}
+
+	return []reconcile.Request{{
+		NamespacedName: types.NamespacedName{
+			Name:      obj.GetName(),
+			Namespace: obj.GetNamespace(),
+		},
+	}}
```
michal-kralik, line 215: This may not be needed, let's check.
Since the `createBackupFunc` creates a `DatabaseClusterBackup` resource, the reconcile loop shall be triggered automatically.

controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +160,206 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+) []reconcile.Request {
+	logger := log.FromContext(ctx)
+	if len(obj.GetOwnerReferences()) == 0 {
+		err := createBackupFunc(ctx, obj)
+		if err != nil {
+			logger.Error(err, fmt.Sprintf("Failed to create DatabaseClusterBackup %s", obj.GetName()))
+		}
+	}
+
+	return []reconcile.Request{{
```
michal-kralik, line 210: I think this request may need to be moved inside the `if` loop if there's no error, right?

controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +160,206 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+		NamespacedName: types.NamespacedName{
+			Name:      obj.GetName(),
+			Namespace: obj.GetNamespace(),
+		},
+	}}
+}
+
+func (r *DatabaseClusterBackupReconciler) tryCreatePG(ctx context.Context, obj client.Object) error {
```
michal-kralik, line 218: We have 3 very similar functions here. It's a good candidate to assess if it would make sense to merge these into one.

controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +160,206 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+
+	backup := &everestv1alpha1.DatabaseClusterBackup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      namespacedName.Name,
+			Namespace: namespacedName.Namespace,
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind:       databaseClusterBackupKind,
+			APIVersion: everestAPIVersion,
+		},
```
michal-kralik, line 241: Since we're using a specific struct `everestv1alpha1.DatabaseClusterBackup`, not `unstructured`, the `TypeMeta` is not required. It shall be added automatically by the k8s client.

controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +160,206 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+	pgBackup := &pgv2.PerconaPGBackup{}
+	namespacedName := types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	}
+
+	if err := r.Get(ctx, namespacedName, pgBackup); err != nil {
+		// if such upstream backup is not found - do nothing
+		if k8serrors.IsNotFound(err) {
+			return nil
```
michal-kralik, line 228: Do we need to check if we shall remove our `DatabaseClusterBackup` in this case?

controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +160,206 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+
+	backup.Spec.DBClusterName = pgBackup.Spec.PGCluster
+
+	cluster := &everestv1alpha1.DatabaseCluster{}
+	err = r.Get(ctx, types.NamespacedName{Name: pgBackup.Spec.PGCluster, Namespace: pgBackup.Namespace}, cluster)
+	if err != nil {
+		return err
```
michal-kralik, line 254: Shall we clean up `DatabaseClusterBackup` in this case?

controllers/databaseclusterbackup_controller.go
```diff
@@ -344,3 +550,17 @@ func (r *DatabaseClusterBackupReconciler) pgRepoName(everestBackup *everestv1alp
...
 	return "", errors.Errorf("can't perform the backup because the requested backup location isn't set in the DatabaseCluster")
 }
+
+func backupStorageName(repoName string, cluster *everestv1alpha1.DatabaseCluster) (string, error) {
+	// repoNames in a PG cluster are in form "repo1", "repo2" etc.
+	// which is mapped to the DatabaseCluster schedules list.
+	// So here we figure out the BackupStorageName of the corresponding schedule.
+	scheduleInd, err := strconv.Atoi(strings.TrimPrefix(repoName, "repo"))
+	if err != nil {
+		return "", errors.New("Unable to get the schedule index")
```
michal-kralik, line 560: Would it make sense to include the `repoName` in the error to make debugging easier in case this fails?

controllers/databaseclusterbackup_controller.go
```diff
@@ -344,3 +550,17 @@ func (r *DatabaseClusterBackupReconciler) pgRepoName(everestBackup *everestv1alp
...
+	}
+	if len(cluster.Spec.Backup.Schedules) < scheduleInd {
+		return "", errors.New("Invalid schedule index")
```
michal-kralik, line 563: Would it make sense to include `scheduleInd` in the error for easier debugging?

controllers/databaseclusterbackup_controller.go
```diff
@@ -226,12 +414,18 @@ func (r *DatabaseClusterBackupReconciler) reconcilePXC(ctx context.Context, back
...
 	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, pxcCR, func() error {
 		pxcCR.TypeMeta = metav1.TypeMeta{
-			APIVersion: pxcBackupAPI,
+			APIVersion: pxcAPIVersion,
 			Kind:       pxcBackupKind,
 		}
 		pxcCR.Spec.PXCCluster = backup.Spec.DBClusterName
 		pxcCR.Spec.StorageName = backup.Spec.BackupStorageName
 
+		pxcCR.SetOwnerReferences([]metav1.OwnerReference{{
```
michal-kralik, line 423: I recommend checking how `BlockOwnerDeletion` is set. Kubernetes says it's `true` by default while `apimachinery` says it's `false`. It's better to check it is configured as we need it.

e2e-tests/tests/dbbackup_pg/03-assert.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: kuttl.dev/v1
+kind: TestAssert
```
michal-kralik, line 2: I recommend adding another step which deletes `DatabaseClusterBackup` and checks if the `PerconaPGBackup` has been delete properly as well.

controllers/databaseclusterbackup_controller.go
recharte: I agree that this isn't needed for the reason you mentioned.

Moreover, I also think this is functionally wrong. We are queuing a request for a `pxcv1.PerconaXtraDBClusterBackup`, `psmdbv1.PerconaServerMongoDBBackup` or `pgv2.PerconaPGBackup` resource and not the `DatabaseClusterBackup` which is definitely not what we want.

controllers/databaseclusterbackup_controller.go
recharte: This is a good point. We actually have a common scenario that illustrates this nicely. We set the retention copies, i.e. number of backups that are kept, and when a new backup exceeds that number the oldest backup will be removed and in that case we need to remove the corresponding DBBackup.
We must have a 1-1 match between DBBackup and the operator's backup CRs.

controllers/databaseclusterbackup_controller.go
recharte: @oksana-grishchenko are you sure that this is actually needed?
It's my understanding that calling `controllerutil.SetControllerReference(backup, pxcCR, r.Client.Scheme())` in line 412 already does this.
AFAIK `SetControllerReference` is basically a `SetOwnerReferences` with the controller flag set to true, so I believe that this section of code actually does nothing.

michal-kralik: It's the second paragraph on this page https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications

controllers/databaseclusterbackup_controller.go
michal-kralik: We spoke on the phone with Oksana and we need to trigger reconciliation here.
My confusion was from the fact I assumed we only need to create the `DatabaseClusterBackup` resource, but we also need to update its status. For this reason it's necessary.

recharte: You guys are correct.

Before, we didn't need the watch because we set the controller reference on the `pxcv1.PerconaXtraDBClusterBackup`, `psmdbv1.PerconaServerMongoDBBackup` and `pgv2.PerconaPGBackup` when we create them and that automatically watches for changes in these resources and triggers a reconciliation of the `DatabaseClusterBackup`.
Now that we explicitly set a watcher, it's overriding the one automatically set when set the controller reference. So we do in fact need to return the list of reconcile requests.

However, I believe that we should return the reference to the `DatabaseClusterBackup` which is what we want to be reconciled and not the `pxcv1.PerconaXtraDBClusterBackup`, `psmdbv1.PerconaServerMongoDBBackup` or `pgv2.PerconaPGBackup` . Right?

oksana-grishchenko: @recharte what do you mean? the `reconcile.Request` contains only the name and the namespace, there is no way to specify the kind of the resource. The controller is created for the `DatabaseClusterBackup` recources and this is the kind that is going to be reconciled. The controller watches the upstream recource, but the reconciling happening with the `DatabaseClusterBackup` only.

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: No, because it defines the objects to be reconciled (as discussed above in [the thread](https://github.com/percona/everest-operator/pull/42#discussion_r1303955597)). Even if we don't need to create a new DatabaseBackup, we still need to reconcile it so it reflects the upstream backup status.

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: I would keep it in separate functions. Even though they are similar to each other right now, each function works with an external operator which may evolve independently and we'll need to adapt it.

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: This is a really good point that have prompted some thoughts and discussions. 

We have this `RetentionCopies` setting which means that at some point the upstream backup object could be deleted by the upstream operator. In this case we need to delete the everest backup CR too. 

I explored the option to set a circular dependency: everest CR owns the upsream CR and vice versa, so when one is deleted, the other one gets deleted automatically. It's not forbidden by any documentation and as it turned out it works well. 
So I added the backward "upstream CR -> everest CR" ownership relation too so we don't need to worry about the deletion.

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: It's not created yet at this point, so there's nothing to be deleted.

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: agree, thanks, updated.

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: Checked the `controllerutil.SetControllerReference` function - sets the `BlockOwnerDeletion` to `true` under the hood, so I added it too. 

@recharte I decided to get rid of the `controllerutil.SetControllerReference` in this controller, bc it used to work well in case we _create_ the upstream backup, but it didn't work well when the upstream backup `was already there`. To avoid two updates on a single object during one method, I decided to use `SetOwnerReference()` instead and have full control over the update.

controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +160,206 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       databaseClusterBackupKind,
+			APIVersion: everestAPIVersion,
+		},
+	}
+
+	err := r.Get(ctx, namespacedName, backup)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
```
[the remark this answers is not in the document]
oksana-grishchenko: thanks, updated

controllers/databaseclusterbackup_controller.go
recharte: Out of curiosity, what do you mean with "didn't work well"?

I'm guessing that it might error out because if the DB operator had already set the controller flag in the owner reference than we wouldn't be able to do the same.
> Since only one OwnerReference can be a controller, it returns an error if there is another OwnerReference with Controller flag set.
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetControllerReference

Was this the case?

controllers/databaseclusterbackup_controller.go
recharte: Yes that's right, I got confused when I saw you were using the name of the upstream backup but you're completely correct, the controller controls the `DatabaseClusterBackup` so that's the resource kind he'll be on the lookout for.
Please disregard my comment, I messed up.

recharte (approved):
Great job Oksana :muscle: :clap:

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: @recharte no, the problem appeared in case the upstream resource already exists and it only needs to update its OwnerReference (it happens when a scheduled backup is performed). 
The `controllerutil.SetControllerReference` din't take any effect, because everything that happens in the later `controllerutil.CreateOrUpdate` doesn't change the CR, so the function returns "unchanged" and the OwnerReference got never updated.

recharte: Makes sense :slightly_smiling_face: thanks for the explaination.

controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +161,223 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+}
+
+func (r *DatabaseClusterBackupReconciler) tryCreatePG(ctx context.Context, obj client.Object) error {
+	pgBackup := &pgv2.PerconaPGBackup{}
+	namespacedName := types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	}
+
+	if err := r.Get(ctx, namespacedName, pgBackup); err != nil {
```
michal-kralik, line 228: I believe we can skip this step and just assert `obj` to `*pgv2.PerconaPGBackup`

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	k8s.io/api v0.28.0
 	k8s.io/apimachinery v0.28.0
 	k8s.io/client-go v0.28.0
+	k8s.io/utils v0.0.0-20230505201702-9f6742963106
```
michal-kralik, line 20: Can we use `github.com/AlekSi/pointer` instead as it's already imported?
The `k8s.io.utils` is also not versioned.

michal-kralik (commented):
Based on reading the code I think we may not be covering a scenario of an on-demand backup being created.
If `DBClusterBackup` is created, from code it seems the upstream backup will be created properly but the `DBClusterBackup` will not have owner reference set properly. Thus if `DBClusterBackup` is deleted, the upstream backup will stay untouched.

If we're expected to have a 1-1 mapping, can we add this as a test to verify?
Once this is resolved, can we add tests to check if `BlockOwnerDeletion` is not going to block removal of either CR due to circular dependency?

controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +161,223 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
 			controller.Owns(&psmdbv1.PerconaServerMongoDBBackup{})
+			controller.Watches(
+				&psmdbv1.PerconaServerMongoDBBackup{},
+				handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []reconcile.Request {
+					return r.tryCreateDBBackups(ctx, obj, r.tryCreatePSMDB)
+				}))
 		}
 	}
 
 	err = r.Get(ctx, types.NamespacedName{Name: pgBackupCRDName}, unstructuredResource)
```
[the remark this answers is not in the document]
oksana-grishchenko: The watcher is added only if the `addPSMDBToScheme()` didn't throw any error which it would if there is no CRD available.

If the CRD installed after we start watching - this is broader problem which is relevant for the DBclusters too. I remember we had a conversation with Diogo about this and he was going to create a ticket for it. @recharte did you have a chance to create it?

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: not sure I understood what do you mean under assert. the structure of two objects is different and this is the way to get the `PerconaPGBackup` resource that corresponds the received object.

go.mod
oksana-grishchenko: thanks for pointing that out! updated

oksana-grishchenko:
@michal-kralik thanks for the comments! Added the deletion tests, which ensure that if the on-demand backup is deleted both the everest and the upstream backups are deleted. The same is for the scheduled backups. 
It works like this because of the OwnerReferences which are set for both resources

controllers/databaseclusterbackup_controller.go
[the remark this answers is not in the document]
michal-kralik: Would the watch fail if the CRD was not in place yet? If not, we could initiate the watch regardless if CRD has been already installed, right?

controllers/databaseclusterbackup_controller.go
michal-kralik: I believe we can do `pgBackup, ok := obj.(*pgv2.PerconaPGBackup)` since the `obj` is coming from a watch for `pgv2.PerconaPGBackup`

controllers/databaseclusterbackup_controller.go
```diff
@@ -153,25 +161,223 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+				handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []reconcile.Request {
+					return r.tryCreateDBBackups(ctx, obj, r.tryCreatePXC)
+				}))
 		}
 	}
 
 	err = r.Get(ctx, types.NamespacedName{Name: psmdbBackupCRDName}, unstructuredResource)
 	if err == nil {
 		if err := r.addPSMDBToScheme(r.Scheme); err == nil {
 			controller.Owns(&psmdbv1.PerconaServerMongoDBBackup{})
```
gen1us2k, line 175: I think that `Owns` here is redundant because `Watches` does almost the same thing here.

Owns tells the manager to trigger the Reconcile(...) whenever any object of <childType> having an ownerReference to <ownerType> is created/updated/deleted.

In fact, For() and Owns() , behind the scenes, are plainly implementations of Watches() only. 

Since Watches will run on any upstream CR event created (create, update, delete) it will trigger `Reconcile` and I think that it's safe to delete

michal-kralik (commented):
Can we please also add a check if the owner references are properly set in the test step `02-assert.yaml`?
We only check if the resources are created, but not if they have owner references set.

oksana-grishchenko:
> Can we please also add a check if the owner references are properly set in the test step 02-assert.yaml?
We only check if the resources are created, but not if they have owner references set.

A good point, sure, the check was there for the on-demand scenario, now it's added for the on-demand backup too. For the on-demand backup owner reference is set only in the direction everest --owns--> upstream. In case the upstream on-demand backup is deleted, it's gonna be recreated.

michal-kralik:
> A good point, sure, the check was there for the on-demand scenario, now it's added for the on-demand backup too. For the on-demand backup owner reference is set only in the direction everest --owns--> upstream. In case the upstream on-demand backup is deleted, it's gonna be recreated.

Is there any reason we don't set the owner references consistently regardless which resource is created?
Why do we want/need to recreate the upstream backup in case it's removed? Isn't this going to trigger another backup?

oksana-grishchenko:
> Is there any reason we don't set the owner references consistently regardless which resource is created?
Why do we want/need to recreate the upstream backup in case it's removed? Isn't this going to trigger another backup?

@michal-kralik Actually, for now we don't cover all the cases like "what if the user goes to k8s and performs that and that action manually", we assume the user only communicates with k8s via Everest. 

Deleting the on-demand upstream backup that was created by an on-demand everest backup - is exactly that manual intervention case. It shouldn't happen, but if it does, the everest will work as expected - it will create (or re-create if it was deleted for some reason) the corresponding upstream backup. 

In case of a scheduled backup the resources creation/deletion is triggered by the upstream operator and it makes sense to manage the resources automatically also, since the user didn't perform any manual action to create/delete that objects.

controllers/databaseclusterbackup_controller.go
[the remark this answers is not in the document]
oksana-grishchenko: we can not initiate the watch for a resource which is not available in the scheme, it will cause an error. 

```
kind must be registered to the Scheme {"error": "no kind is registered for the type v1.PerconaServerMongoDBBackup in scheme \"pkg/runtime/scheme.go:100\""}
```

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: No, the obj that is passed here has a different structure, [here](https://go.dev/play/p/iHAMl6V1tOE) is an example of what it could be

e2e-tests/tests/dbbackup_pg/03-assert.yaml
oksana-grishchenko: The deletion tests are added

PR #45, opened 2023-08-23
EVEREST-296 allow on-demand backups without defined schedules

recharte:
EVEREST-296
In order for us to allow users to create on-demand backups without having a backup schedule defined, we ensure that the referenced backup storage is available in the underlying DB CR at the DBBackup CR reconciliation.
Also, we make sure to also create/update this backup storage when reconciling the DBCluster CR.

Note for reviewers: I added tests for PXC and PSMDB but I'm still missing the PG tests for this feature. I'll add them shortly but I decided to publish the PR regardless so you can start taking a look at it.


controllers/databasecluster_controller.go
```diff
@@ -291,7 +293,55 @@ func (r *DatabaseClusterReconciler) genPSMDBBackupSpec(
...
+	// If scheduled backups are disabled, just return the storages used in
+	// DatabaseClusterBackup objects
+	if !database.Spec.Backup.Enabled {
+		psmdbBackupSpec.Storages = storages
+		return psmdbBackupSpec, nil
+	}
+
 	var tasks []psmdbv1.BackupTaskSpec //nolint:prealloc
 	for _, schedule := range database.Spec.Backup.Schedules {
 		if !schedule.Enabled {
```
michal-kralik, line 347: Here we can also check if `storages[schedule.BackupStorageName]` is already set and skip this storage.

controllers/databasecluster_controller.go
```diff
@@ -607,11 +655,11 @@ func (r *DatabaseClusterReconciler) getSecretFromMonitoringConfig(
 	return secretData, nil
 }
 
-// createOrUpdateSecret creates or updates a secret by its name.
+// CreateOrUpdateSecret creates or updates a secret by its name.
 // If a secret is created and secretName is empty, generateName is used
 // to generate a unique name.
-func (r *DatabaseClusterReconciler) createOrUpdateSecret(
-	ctx context.Context, database *everestv1alpha1.DatabaseCluster,
+func CreateOrUpdateSecret(
```
michal-kralik, line 661: I suggest we extract the methods which are no longer in `DatabaseClusterReconciler` into a separate file so it's clear they can be reused.

controllers/databasecluster_controller.go
```diff
@@ -1050,45 +1146,129 @@ func (r *DatabaseClusterReconciler) reconcilePXC(ctx context.Context, req ctrl.R
 	return r.Status().Update(ctx, database)
 }
 
-func (r *DatabaseClusterReconciler) createPGBackrestSecret(
+// Creates or updates the PG Backrest secret
+func CreatePGBackrestSecret(
```
michal-kralik, line 1150: Is there any reason the functions removed from `DatabaseClusterReconciler` are now exported? It doesn't seem they need to be, right?

controllers/databasecluster_controller.go
```diff
@@ -1050,45 +1146,129 @@ func (r *DatabaseClusterReconciler) reconcilePXC(ctx context.Context, req ctrl.R
...
+	return -1
+}
+
+// Generates a PGBackrest repo for a given backup storage
+func GenPGBackrestRepo(
+	repoIdx int,
+	backupStorage *everestv1alpha1.BackupStorage,
+	backupSchedule *string,
+) (crunchyv1beta1.PGBackRestRepo, error) {
+	if repoIdx < 0 || repoIdx > 4 {
```
michal-kralik, line 1203: Is there any reason the indexes are limited to 0-4?

controllers/databasecluster_controller.go
```diff
@@ -1124,46 +1318,99 @@ func (r *DatabaseClusterReconciler) genPGBackupsSpec(
...
+
+		repoIdx := GetBackupStorageIndexInPGBackrestRepo(backupStorage, repos)
+		// Already exists in the list of repos so skip
+		if repoIdx != -1 {
+			continue
+		}
+
+		repoIdx = len(repos)
+		if repoIdx >= 4 {
+			return crunchyv1beta1.Backups{}, errors.Errorf("number of backup repos for postgresql cannot exceed 4")
```
michal-kralik, line 1352: Based on line 1203 it seems the limit is 0-4 which is 5.

controllers/databasecluster_controller.go
```diff
@@ -1124,46 +1318,99 @@ func (r *DatabaseClusterReconciler) genPGBackupsSpec(
...
+		}
+
+		pgRepo, err := GenPGBackrestRepo(repoIdx+1, backupStorage, nil)
```
michal-kralik, line 1355: Is it correct to reference `repoIdx+1`. Shouldn't it be just `repoIdx`?

controllers/databasecluster_controller.go
```diff
@@ -1050,45 +1146,129 @@ func (r *DatabaseClusterReconciler) reconcilePXC(ctx context.Context, req ctrl.R
...
+		fmt.Sprintf("%s-s3-key-secret", repoName),
+		string(s3StorageSecret.Data["AWS_SECRET_ACCESS_KEY"]),
+	)
+	if err != nil {
+		return data, err
+	}
+
+	buf := new(bytes.Buffer)
+	cfg.WriteTo(buf)
+	return buf.Bytes(), nil
```
michal-kralik, line 1269: This seems to be an internal function.
Can we, instead of passing around `[]byte`, return `*ini.File` and marshal this to `[]byte`` only when needed in the spec?

This will save us from loading/writing bytes.

controllers/databaseclusterbackup_controller.go
```diff
@@ -221,10 +242,55 @@ func (r *DatabaseClusterBackupReconciler) reconcilePXC(ctx context.Context, back
 			Namespace: backup.Namespace,
 		},
 	}
+
+	// Given the more imperative-like nature of the DatabaseClusterBackup we
+	// can't afford to wait for the DatabaseCluster to be reconciled to have
+	// the necessary storage defined because this could lead to a failure and
+	// the backup wouldn't be retried. Instead, we check and update the
+	// storages map here to make sure the PerconaXtraDBCluster has the
+	// necessary storage defined to be able to perform the backup.
```
michal-kralik, line 251: Could this approach lead to race-conditions?
Such as a db cluster is being reconciled, but a backup is triggered and we overwrite dbcluster values or fail because of the revision number not matching?

Wouldn't it make more sense to keep this in the db cluster reconciliation loop and retry backup reconciliation until the storages are properly defined?

michal-kralik (commented):
Great that you shared this early Diogo.
Left a couple comments.

controllers/databasecluster_controller.go
recharte: I also thought about that and agree that it's the best approach going forward. I'll make that change.

controllers/databasecluster_controller.go
recharte: I removed the `DatabaseClusterReconciler` and exported them specifically because they are needed in the `DatabaseClusterBackupReconciler` as well.

controllers/databasecluster_controller.go
recharte: This is a limitation of the PG operator. I don't know if there's any reason for that limitation on their side or not.

recharte: This just made me notice that we have a bug. The repo name must be repo[1-4] and we're allowing 0 as an input which isn't correct.

controllers/databasecluster_controller.go
michal-kralik: Yes, they're in the same package. No need to export the methods.

controllers/databasecluster_controller.go
michal-kralik: I'm a bit worried about the 4 repos limit.
It can easily happen there will be more than 4 repos defined by the user in which case we automatically fail the reconcile loop and cannot make any more changes to the `DBCluster`.

Maybe we may need to handle this gracefully?
Or maybe create the repos only if they're really needed - e.g. when a backup is triggered?
Or maybe impose a limit in the backend so we prevent the users from having more than 4 repos?

What do you think?

controllers/databasecluster_controller.go
recharte: Actually we can't, when processing the schedules we not only ensure the storage is set but also add a task to run the backup on schedule (right at the end of the for loop).

controllers/databasecluster_controller.go
recharte: Yes, you're correct, thanks for spotting it, I'll fix it

controllers/databasecluster_controller.go
recharte: Yes, I completely agree with your points and have those concerns myself.
I actually wanted to have this as part of the extra validations we were planning on adding in the near future. I wanted to do this with admission webhooks like we explored in #34.
However Andrew shared this feedback about the webhooks so it's not entirely clear what path we will follow.
> Well. There are a couple of drawbacks to using webhooks:
>
> It makes the operator cluster wide which is a deal breaker for upstream operator customers and Peter Szczepaniak told that we will hit the same wall sooner or later 
> It requires an effort to improve CI/CD pipeline and it will make it even more unstable because webhooks are cluster-wide
> We need to enable deployments of an operator image into the EKS cluster for our pipeline which requires a lot of work to stabilize tests 
> We can face other unknown issues with our pipeline. 
> As an option, we can disable webhooks for pipeline runs yet we need to generate tls certificates anyway to make operator work

recharte: So I would like to go with a 4th option which is pretty similar to your 3rd one but doing it in the operator instead by using admission webhooks.
i.e. impose a limit in the operator so we prevent the users from having more than 4 repos

If this isn't possible we'll have to go with the BE validation approach.

oksana-grishchenko (commented):
Since the PG is the trickiest part, lets make sure that it works properly by adding the tests.
In general it looks fine, thanks for implementing it! 🙌

controllers/databaseclusterbackup_controller.go
recharte: > Wouldn't it make more sense to keep this in the db cluster reconciliation loop and retry backup reconciliation until the storages are properly defined?

Initially, I thought this wouldn't work because I was always thinking about immediately creating the PerconaXtraDBClusterBackup (before the DBC reconciliation had the opportunity to add the storages) and that would lead to a that backup request failing for ever, backup reconciliation wouldn't fix it.

Now I'm thinking that, if the storage isn't yet available, we can fail the backup reconciliation without creating the PerconaXtraDBClusterBackup. Then, the DBC reconciliation will add the storages and when the DBB reconcilation is run afterwards it will succeed and this should work.

I like this approach much better, I'll try to implement it and let you know. Thank you for questioning this and making me reevaluate my initial assumptions.

controllers/databasecluster_controller.go
michal-kralik: Wouldn't this put us in the same situation as we're now - i.e. everest operator will fail to reconcile the db cluster and no changes to the cluster will be possible until the too many repos issue is resolved?

recharte: No, the admission webhooks are run before storing the CR so if it fails the CR won't be changed. Think of it as a validation before committing a change to a DB, if the validation fails nothing gets changed.

controllers/databaseclusterbackup_controller.go
gen1us2k: We can introduce a new state here and use [ensureClusterIsReady](https://github.com/percona/everest-operator/blob/main/controllers/databaseclusterrestore_controller.go#L97-L115) that we have for restores

gen1us2k (commented):
Just out of curiosity. How it will change for the API users? Should they fill the same structs for the backup section but without schedules being set up?

recharte:
Hi guys, I reworked the way we generate the PG repos and added a bunch of tests for it, it should now be less error.

I also removed the reconciliation in the DatabaseClusterBackup controller. Now, if the storage isn't present in the PXC/PSMDB/PG cluster CR we requeue the DatabaseClusterBackup reconciliation for later and the DatabaseCluster controller is the only controller responsible for reconciling those storages. This should avoid any race-conditions and also reduces the DatabaseClusterBackup controller complexity.

gen1us2k:
Yeah. Sorry for the unclear question. I mean that during the creation of a database cluster we need to have a filled `backup.Schedules` section in the databaseCluster CR. As a user who wants to use only on-demand backups I still need to specify backupStorageName somehow. Do I need to fill in the same Schedules field? In that case, it would be a bit misleading from API user standpoint

controllers/databaseclusterbackup_controller.go
```diff
@@ -288,29 +399,99 @@ func (r *DatabaseClusterBackupReconciler) reconcilePSMDB(ctx context.Context, ba
...
+	if repoIdx == -1 {
+		// Replace the default PVC repo if it exists
+		if len(pgDBCR.Spec.Backups.PGBackRest.Repos) == 1 && pgDBCR.Spec.Backups.PGBackRest.Repos[0].Volume != nil {
+			repoIdx = 0
+		} else {
+			repoIdx = len(pgDBCR.Spec.Backups.PGBackRest.Repos)
+		}
+		if repoIdx >= 4 {
+			return errors.Errorf("number of backup repos for postgresql cannot exceed 4")
+		}
```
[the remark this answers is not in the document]
recharte: Either that or the proposal presented [here](https://github.com/percona/everest-operator/pull/45#discussion_r1305454122) it also provides instant feedback to the user.

recharte:
> Yeah. Sorry for the unclear question. I mean that during the creation of a database cluster we need to have a filled `backup.Schedules` section in the databaseCluster CR. As a user who wants to use only on-demand backups I still need to specify backupStorageName somehow. Do I need to fill in the same Schedules field? In that case, it would be a bit misleading from API user standpoint

Yeah, the whole point of this PR is precisely to avoid having to fill the `backup.Schedules` section in the databaseCluster CR. With the changes in this PR, the user may request on-demand backups regardless of having set the `backup.Schedules` section.
As a user who wants to use only on-demand backups I can now create a DatabaseClusterBackup CR without having `backup.Schedules` defined.

gen1us2k (approved):
Great implementation. Thanks for that. Good job!

oksana-grishchenko (approved):
Great job, Diogo!🚀  Thanks for implementing unit-tests for the tricky PG part, it brings confidence that it works as expected. 

One question. Could it happen that we create a backup with a not yet included BackupStorage, but it will never succeed bc the DatabaseCluster reconciliation loop was never triggered since?

recharte:
> One question. Could it happen that we create a backup with a not yet included BackupStorage, but it will never succeed bc the DatabaseCluster reconciliation loop was never triggered since?

I think that this should be something we add in the validations that we've been discussing. We shouldn't allow a backup to be created for BackupStorage that doesn't exist.

recharte:
Hey guys, I just force pushed the rebase with main to fix the conflicts.

e2e-tests/tests/dbbackup_pg/03-assert.yaml
```diff
@@ -1,10 +1,51 @@
...
 commands:
   # check that there is no dbb/pg-backup CRs left
   - script: >
-      test "$(kubectl get dbb -n $NAMESPACE -o jsonpath='{.items}')" == "[]"
+      test "$(kubectl get dbb -n $NAMESPACE -o jsonpath='{.items}')" = "[]"
   - script: >
-      test "$(kubectl get pg-backup -n $NAMESPACE -o jsonpath='{.items}')" == "[]"    
+      test "$(kubectl get pg-backup -n $NAMESPACE -o jsonpath='{.items}')" = "[]"
 
+timeout: 400
```
oksana-grishchenko, line 11: `timeout` key is duplicated

recharte: thanks, fixed

PR #50, opened 2023-08-28
EVEREST-256 DatabaseCluster CR validation improvements

gen1us2k:
It uses https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/ which limits supported versions of k8s from 1.25. However, 

```
1.25 
Latest Release:1.25.13 (released: 2023-08-23)
End of Life:2023-10-28
Patch Releases:
```


api/v1alpha1/databasecluster_types_test.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			e := Expose{}
+			if got := e.toCIDR(tt.ranges); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("Expose.parseIPSourceRanges() = %v, want %v", got, tt.want)
```
michal-kralik, line 53: Can we please update the text in the error to reflect the new function?

.github/workflows/ci.yml
```diff
@@ -154,15 +154,36 @@ jobs:
         env:
           CLUSTER_NAME: "dbaas-ci-20221005"
           NODEPOOL_NAME: "terraform-node-dbaas-ci-20221005"
+      - name: Clean up custom resources
+        if: always()
+        run: |
+          kubectl delete db --all-namespaces --all
+          kubectl delete pvc --all-namespaces --all
```
michal-kralik, line 161: Wouldn't this affect all other tests running in parallel?

oksana-grishchenko: I'm also concerned about the parallel runs. kuttl allows to set the namespace name instead of using the generated one, so I'm wondering if we could use it somehow to distinguish what namespaces to clean up?

oksana-grishchenko (approved):
The validation looks good, thanks for the implementaion. One concern regarding the cleanup, posted a comment. 

Approving the request since the main purpose of this PR is to improve validation. Feel free to address the cleanup problem in another PR.

.github/workflows/ci.yml
gen1us2k: No. It runs only once because of concurrency setting

gen1us2k: https://github.com/percona/everest-operator/blob/b09599c577495224c8f80ef37a861ce53c441587/.github/workflows/ci.yml#L53 check this. It runs only once and the second run won't be triggered until the previous one is finished.

PR #51, opened 2023-08-30
EVEREST-249 restores redesign

recharte:
EVEREST-249

Sorry for the lack of a proper description, too tired tonight.
You folks know what this is about, redesign the restores API for both new DBCs using the DataSource field as well as existing DBC using the DatabaseClusterRestore CRs.
The commits should be relatively easy to follow :slightly_smiling_face:


api/v1alpha1/databasecluster_types.go
```diff
@@ -145,13 +145,20 @@ type Proxy struct {
 	Resources Resources `json:"resources,omitempty"`
 }
 
+// BackupSource represents settings of a source where to get a backup to run restoration.
+type BackupSource struct {
+	// Path is the path to the backup file/directory.
+	Path string `json:"path"`
```
gen1us2k, line 151: Is it the same thing as the destination on upstream operators?

api/v1alpha1/databasecluster_types.go
```diff
@@ -145,13 +145,20 @@ type Proxy struct {
...
+
 // DataSource is the data source configuration.
 type DataSource struct {
-	// BackupName is the name of the backup from backup location to use
-	BackupName string `json:"backupName"`
-	// BackupStorageName is the name of the BackupStorage CR that defines the
-	// storage location
-	BackupStorageName string `json:"backupStorageName"`
+	// DBClusterBackupName is the name of the DB cluster backup to restore from
+	DBClusterBackupName string `json:"dbClusterBackupName,omitempty"`
```
gen1us2k, line 159: Do you think that dbClusterBackupName is good enough? What about clusterBackupName? It's part of DatabaseCluster CR and databaseCluster.spec.dataSource.clusterBackupName sounds clear while databaseCluster.spec.dataSource.dbClusterBackupName sounds a bit redundant. What do you think?

api/v1alpha1/databaseclusterrestore_types.go
```diff
@@ -30,49 +30,32 @@ const (
...
-		Image                 string                     `json:"image,omitempty"`
-		SSLSecretName         string                     `json:"sslSecretName,omitempty"`
-		SSLInternalSecretName string                     `json:"sslInternalSecretName,omitempty"`
-		VaultSecretName       string                     `json:"vaultSecretName,omitempty"`
-	}
-)
+// DatabaseClusterRestoreSpec defines the desired state of DatabaseClusterRestore.
+type DatabaseClusterRestoreSpec struct {
+	// DBClusterName defines the cluster name to restore.
+	DBClusterName string `json:"dbClusterName"`
```
gen1us2k, line 39: It's the same story here that clusterName sounds clear enough because restores can be applied against db clusters :). However, I'm okay with the proposed naming

gen1us2k (commented):
Thanks for the contribution. It looks good so far but I left minor comments

api/v1alpha1/databaseclusterrestore_types.go
recharte: I only called it like this because it was the name we already used in the DatabaseClusterBackup and wanted to keep it consistent.

api/v1alpha1/databasecluster_types.go
recharte: I went with this approach to keep it consistent with the `DBClusterName` we already had in the `DatabaseClusterBackup`. My reasoning was if a reference to the `DatabaseCluster` is `DBClusterName` then a reference to a `DatabaseClusterBackup` should be called `DBClusterBackupName`.
I can change this if you and @oksana-grishchenko think it's better but in that case I'd also change the  `DBClusterName` we already had in the `DatabaseClusterBackup` for consistency.

gen1us2k: Let's create a ticket for that and keep it as is

gen1us2k (approved):

controllers/databaseclusterrestore_controller.go
```diff
@@ -78,32 +79,59 @@ func (r *DatabaseClusterRestoreReconciler) Reconcile(ctx context.Context, req ct
...
-			logger.Error(err, "unable to restore PXC Cluster")
+	if dbCR.Spec.Engine.Type == everestv1alpha1.DatabaseEnginePSMDB {
+		if err := r.restorePSMDB(ctx, cr); err != nil {
+			// The DatabaseCluster controller is responsible for updating the
+			// upstream DB cluster with the necessary storage definition. If
+			// the storage is not defined in the upstream DB cluster CR, we
+			// requeue the backup to give the DatabaseCluster controller a
+			// chance to update the upstream DB cluster CR.
+			if errors.Is(err, ErrBackupStorageUndefined) {
+				return reconcile.Result{Requeue: true}, nil
```
oksana-grishchenko, line 114: why only the psmdb gives another chance to reconcile but not the other operators?

bundle/manifests/everest.percona.com_databaseclusterrestores.yaml
```diff
@@ -158,86 +102,11 @@ spec:
...
-                items:
-                  description: "Condition contains details for one aspect of the current
-                    state of this API Resource. --- This struct is intended for direct
-                    use as an array at the field path .status.conditions.  For example,
-                    \n type FooStatus struct{ // Represents the observations of a
-                    foo's current state. // Known .status.conditions.type are: \"Available\",
-                    \"Progressing\", and \"Degraded\" // +patchMergeKey=type // +patchStrategy=merge
-                    // +listType=map // +listMapKey=type Conditions []metav1.Condition
-                    `json:\"conditions,omitempty\" patchStrategy:\"merge\" patchMergeKey:\"type\"
-                    protobuf:\"bytes,1,rep,name=conditions\"` \n // other fields }"
```
oksana-grishchenko, line 171: so it was like a "nice to have" which wasn't implemented anyway, so we deleted it?

oksana-grishchenko (approved):
Looks much cleaner it was before, great job! Left couple of comments just to understand better the decisions made.

controllers/databaseclusterbackup_controller.go
```diff
@@ -554,10 +561,71 @@ func (r *DatabaseClusterBackupReconciler) reconcilePSMDB(ctx context.Context, ba
...
+	logger := log.FromContext(ctx)
+	backupStorageSecret := &corev1.Secret{}
+	err := r.Get(ctx, types.NamespacedName{Name: backupStorage.Spec.CredentialsSecretName, Namespace: backupStorage.Namespace}, backupStorageSecret)
+	if err != nil {
+		logger.Error(err, "unable to get backup storage secret")
+		return nil
+	}
+
+	cfg, err := config.LoadDefaultConfig(ctx,
+		config.WithRegion("eu-west-3"),
```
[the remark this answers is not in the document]
recharte: My bad :sweat_smile: thanks for spotting it

controllers/databaseclusterrestore_controller.go
recharte: The other operators don't need to have the storage defined in their CRs for the restore process.
They only require it for the backups.

bundle/manifests/everest.percona.com_databaseclusterrestores.yaml
recharte: Actually we didn't even set this anywhere. I don't know what was the initial intention for adding it to the API but it wasn't used anywhere.

controllers/databasecluster_controller.go
```diff
@@ -2009,8 +2097,8 @@ func (r *DatabaseClusterReconciler) reconcileLabels(ctx context.Context, databas
 		needsUpdate = true
 	}
 	if database.Spec.DataSource != nil {
-		if _, ok := database.ObjectMeta.Labels[fmt.Sprintf(backupStorageNameLabelTmpl, database.Spec.DataSource.BackupStorageName)]; !ok {
-			database.ObjectMeta.Labels[fmt.Sprintf(backupStorageNameLabelTmpl, database.Spec.DataSource.BackupStorageName)] = backupStorageLabelValue
+		if _, ok := database.ObjectMeta.Labels[fmt.Sprintf(backupStorageNameLabelTmpl, database.Spec.DataSource.DBClusterBackupName)]; !ok {
+			database.ObjectMeta.Labels[fmt.Sprintf(backupStorageNameLabelTmpl, database.Spec.DataSource.DBClusterBackupName)] = backupStorageLabelValue
```
oksana-grishchenko, line 2101: Posting a comment hoping to figure out if the change served to a purpose or it was unintended.

recharte: ohhh boy, over 1 year old PR :sweat_smile:
this was done in a commit called "[Fix labels](https://github.com/percona/everest-operator/pull/51/commits/363113d3c9c29af898ea73dba5f2726e86081af8)" so it was definitely intentional but I can't recall why and now it just feels wrong :see_no_evil:

gen1us2k: Iirc it was done for faster filtering backups/restores per db cluster because of k8s API capabilities to filter by labels

gen1us2k: So. After digging it a bit deeper I remembered the following stuff:
1. It was done in the first place for faster filtering of database clusters to delete/update backups on the everest backend
2. During the previous review we forgot to use `Spec.DataSource.BackupStorageName` instead of `Spec.DataSource.DBClusterBackupName` which is a mistake here
3. On the everest-backend there was a check and the backend was fetching db clusters from k8s using labels for [faster search](https://github.com/percona/percona-everest-backend/blob/main/pkg/kubernetes/backup_storage.go#L58)
4. However, this use of labels was gone during the move to a monorepo or to something else. You can take a look here https://github.com/percona/everest/blob/main/pkg/kubernetes/backup_storage.go#L63. Using labels was intended for faster/better way of finding db clusters that use a backup storage without looping through all database clusters
I hope this sheds some light on  the situation

oksana-grishchenko: Thanks @gen1us2k! Created an overhead ticket to review & improve the labelling policy EVEREST-1654

PR #57, opened 2023-09-01
EVEREST-324 Fixed scaling for pxc and psmdb

gen1us2k:
EVEREST-324

Dropped sharded mongo because it should be done in a different way and we received feedback to have this option in the UI. Until then we do not support sharded mongo instances.
Toggling unsafeConfigurations from true to false and vice versa may lead to unexpected behavior in operator

### DIFF
--- a/api/v1alpha1/databasecluster_types.go
+++ b/api/v1alpha1/databasecluster_types.go
@@ -197,7 +197,7 @@ type DatabaseClusterSpec struct {
 	// Paused is a flag to stop the cluster
 	Paused bool `json:"paused,omitempty"`
 	// AllowUnsafeConfiguration field used to ensure that the user can create configurations unfit for production use.
-	AllowUnsafeConfiguration bool `json:"unsafeConfiguration,omitempty"`
+	AllowUnsafeConfiguration bool `json:"allowUnsafeConfiguration,omitempty"`
 	// Engine is the database engine specification
 	Engine Engine `json:"engine"`
 	// Proxy is the proxy specification. If not set, an appropriate
--- a/api/v1alpha1/databasecluster_types.go
+++ b/api/v1alpha1/databasecluster_types.go
@@ -97,7 +97,7 @@ type Resources struct {
 type Engine struct {
 	// Type is the engine type
 	Type EngineType `json:"type"`
-	// UnsafeConfiguration field used to ensure bla bla bla
+	// UnsafeConfiguration field used to ensure that the user can create configurations unfit for production use.
 	UnsafeConfiguration bool `json:"unsafeConfiguration,omitempty"`
 	// Version is the engine version
 	Version string `json:"version,omitempty"`
--- a/api/v1alpha1/databasecluster_types.go
+++ b/api/v1alpha1/databasecluster_types.go
@@ -97,6 +97,8 @@ type Resources struct {
 type Engine struct {
 	// Type is the engine type
 	Type EngineType `json:"type"`
+	// UnsafeConfiguration field used to ensure bla bla bla
+	UnsafeConfiguration bool `json:"unsafeConfiguration,omitempty"`
 	// Version is the engine version
 	Version string `json:"version,omitempty"`
 	// Replicas is the number of engine replicas
--- a/api/v1alpha1/databasecluster_types.go
+++ b/api/v1alpha1/databasecluster_types.go
@@ -97,8 +97,6 @@ type Resources struct {
 type Engine struct {
 	// Type is the engine type
 	Type EngineType `json:"type"`
-	// UnsafeConfiguration field used to ensure that the user can create configurations unfit for production use.
-	UnsafeConfiguration bool `json:"unsafeConfiguration,omitempty"`
 	// Version is the engine version
 	Version string `json:"version,omitempty"`
 	// Replicas is the number of engine replicas
@@ -198,6 +196,8 @@ type Monitoring struct {
 type DatabaseClusterSpec struct {
 	// Paused is a flag to stop the cluster
 	Paused bool `json:"paused,omitempty"`
+	// AllowUnsafeConfiguration field used to ensure that the user can create configurations unfit for production use.
+	AllowUnsafeConfiguration bool `json:"unsafeConfiguration,omitempty"`
 	// Engine is the database engine specification
 	Engine Engine `json:"engine"`
 	// Proxy is the proxy specification. If not set, an appropriate
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.12
+VERSION ?= 0.0.13
 
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "candidate,fast,stable")
--- a/Makefile
+++ b/Makefile
@@ -309,3 +309,6 @@ DEPLOYDIR = ./deploy
 .PHONY: $(DEPLOYDIR)/bundle.yaml
 $(DEPLOYDIR)/bundle.yaml: manifests kustomize ## Generate deploy/bundle.yaml
 	$(KUSTOMIZE) build config/default -o $(DEPLOYDIR)/bundle.yaml
+
+.PHONY: release
+release:  generate manifests bundle format $(DEPLOYDIR)/bundle.yaml ## Prepare release
--- a/bundle/manifests/everest-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/everest-operator.clusterserviceversion.yaml
@@ -103,10 +103,10 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2023-09-01T13:01:35Z"
+    createdAt: "2023-09-01T13:57:46Z"
     operators.operatorframework.io/builder: operator-sdk-v1.31.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
-  name: everest-operator.v0.0.12
+  name: everest-operator.v0.0.13
   namespace: placeholder
 spec:
   apiservicedefinitions: {}
@@ -512,7 +512,7 @@ spec:
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.annotations['olm.targetNamespaces']
-                image: docker.io/percona/everest-operator:0.0.12
+                image: docker.io/percona/everest-operator:0.0.13
                 livenessProbe:
                   httpGet:
                     path: /healthz
@@ -606,4 +606,4 @@ spec:
   provider:
     name: Percona
     url: https://percona.com
-  version: 0.0.12
+  version: 0.0.13
--- a/bundle/manifests/everest-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/everest-operator.clusterserviceversion.yaml
@@ -103,7 +103,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2023-09-01T13:57:46Z"
+    createdAt: "2023-09-01T14:02:56Z"
     operators.operatorframework.io/builder: operator-sdk-v1.31.0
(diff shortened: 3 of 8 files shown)


api/v1alpha1/databasecluster_types.go
```diff
@@ -97,6 +97,8 @@ type Resources struct {
 type Engine struct {
 	// Type is the engine type
 	Type EngineType `json:"type"`
+	// UnsafeConfiguration field used to ensure bla bla bla
```
recharte, line 100:
```suggestion
	// UnsafeConfiguration field used to ensure that the user can create configurations unfit for production use.
```

api/v1alpha1/databasecluster_types.go
```diff
@@ -97,6 +97,8 @@ type Resources struct {
...
+	UnsafeConfiguration bool `json:"unsafeConfiguration,omitempty"`
```
recharte, line 101: What do you think of putting this outside of the Engine? We might have "unsafe configs" for other components of the DBC CR.

recharte: Also, I would name it `AllowUnsafeConfiguration` because setting this to true doesn't make the configuration inherently unsafe, you just allow for unsafe configs to be defined.

api/v1alpha1/databasecluster_types.go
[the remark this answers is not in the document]
recharte: Best use of this linter I've ever seen :joy:

recharte:
I'm a bit worried about what this solution means in terms of effort for the FE. To properly handle it we need to know when we need to set the UnsafeConfiguration field.. For now we can set it to always be true but in that case we could also do that in the operator itself and not expose it.
What are your thought on this @gen1us2k?

api/v1alpha1/databasecluster_types.go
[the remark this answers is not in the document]
gen1us2k: Yeah :D

gen1us2k:
> For now, we can set it to always be true
it's true if a user started with a single node cluster. For a 3-node cluster and above it will be false.

api/v1alpha1/databasecluster_types.go
gen1us2k: Done. Thanks for catching this

api/v1alpha1/databasecluster_types.go
```diff
@@ -196,6 +196,8 @@ type Monitoring struct {
 type DatabaseClusterSpec struct {
 	// Paused is a flag to stop the cluster
 	Paused bool `json:"paused,omitempty"`
+	// AllowUnsafeConfiguration field used to ensure that the user can create configurations unfit for production use.
+	AllowUnsafeConfiguration bool `json:"unsafeConfiguration,omitempty"`
```
recharte, line 200:
```suggestion
	AllowUnsafeConfiguration bool `json:"allowUnsafeConfiguration,omitempty"`
```

recharte:
> > For now, we can set it to always be true
> 
> it's true if a user started with a single node cluster. For a 3-node cluster and above it will be false.

How are you planning on implementing this immutability. Should the FE or the BE be responsible for this?

recharte:
> > I'm a bit worried about what this solution means in terms of effort for the everest FE/BE. To properly handle it we need to know when to set the UnsafeConfiguration field. For now we can set it to always be true but in that case we could also do that in the operator itself without expose that field and changing the API.
> 
> Well, we need to set unsafeConfiguration for the upstream operators only to support single Node clusters (at least for now). There are other cases for it as well, but I'm not sure that we will need to change it from FE/BE or other places.
> 
> Immutability for unsafe options is made for now because not all upstream operators support it. AFAIK it worked perfectly for PXC a while ago but right now it's either broken or I'm mistaken here. PSMDB does not support it and it should be immutable and I don't know anything about PG right now.
> 
> I thought about adding it without affecting CR but we need to persist it somewhere and etcd looks like a great solution, hence a new field.
> 
> I'll update backend specs to make it consistent but usually we do not need to set this field. Or maybe later there will be cases to enable it.

The thing is that with this change you don't guarantee immutability you just pass on that responsibility to the operator's user, which for the full-stack everest will be either the FE or the BE. So who will guarantee that this field won't be changed by updates?

oksana-grishchenko (approved):
If we really need the new field in CR, let's make the BE responsible for preventing the `AllowUnsafeConfiguration` field changes and add smth like [this check](https://github.com/percona/percona-everest-backend/blob/0f725221ce995d15bbbe730d18ab5651c8fd5ee3/api/database_cluster.go#L140-L148) for it during the database cluster update

recharte (approved):
I don't think this is the best approach because it gives the user a false sense of control security for something which we could handle ourselves in the operator without exposing it to the user.

We are vulnerable to weird cases such as when the users set this
```
database.Spec.Engine.Replicas = 1
database.Spec.AllowUnsafeConfiguration = false
```
Where they explicitly declare that they don't want to allow unsafe configurations. But we just go like "Actually you do, you want 1 replica, here you go use an usafe config, we even changed it for you" :joy: 
```
database.Spec.AllowUnsafeConfiguration = true
```

My point is that if we want to make sure that the AllowUnsafeConfiguration we pass along to the DB operators is immutable we could do it internally in a more consistent manner without exposing it to the users.

Regardless, this approach fixes the issue at hand so I'll unblock the PR.

recharte:
> if we want to make sure that the AllowUnsafeConfiguration we pass along to the DB operators is immutable we could do it internally in a more consistent manner without exposing it to the users

The current implementation doesn't guarantee immutability even if the user doesn't change it explicitally.
e.g. if the users set
```
database.Spec.Engine.Replicas = 3
```
and then change it to 
```
database.Spec.Engine.Replicas = 1
```
The AllowUnsafeConfiguration will also move from `false` to `true`.


I believe that if we don't want users to be able to change it and we actually infer it from other fields regardless, we don't need to have it the spec.
If we really want to make it available for the users to see that they're using an unsafe config maybe we should expose it in some other place like an annotation or the status.

recharte:
I haven't tested it but I think that the following code would actually guarantee the immutability of the `psmdb.Spec.UnsafeConf ` without exposing anything to the user.
```
	psmdbCR := &everestv1alpha1.DatabaseEngine{}
	psmdbCRNotFound := k8serrors.IsNotFound(r.Get(ctx, types.NamespacedName{Namespace: database.Namespace, Name: database.Name}, psmdbCR))

	if err := controllerutil.SetControllerReference(database, psmdb, r.Client.Scheme()); err != nil {
		return err
	}
	_, err = controllerutil.CreateOrUpdate(ctx, r.Client, psmdb, func() error {
		if psmdbCRNotFound {
			psmdb.Spec.UnsafeConf = database.Spec.Engine.Replicas == 1
		}
```

recharte:
I haven't tested it but I think that the following diff would actually guarantee the immutability of the `psmdb.Spec.UnsafeConf ` without exposing anything to the user.
```
diff --git a/controllers/databasecluster_controller.go b/controllers/databasecluster_controller.go
index d58165f..4fa53be 100644
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -458,6 +458,9 @@ func (r *DatabaseClusterReconciler) reconcilePSMDB(ctx context.Context, req ctrl
                return errors.Wrapf(err, "failed to get database engine %s", operatorDeployment[database.Spec.Engine.Type])
        }

+       psmdbCR := &psmdbv1.PerconaServerMongoDB{}
+       psmdbCRNotFound := k8serrors.IsNotFound(r.Get(ctx, types.NamespacedName{Namespace: database.Namespace, Name: database.Name}, psmdbCR))
+
        if err := controllerutil.SetControllerReference(database, psmdb, r.Client.Scheme()); err != nil {
                return err
        }
@@ -502,7 +505,11 @@ func (r *DatabaseClusterReconciler) reconcilePSMDB(ctx context.Context, req ctrl
                }

                psmdb.Spec.CRVersion = version.ToCRVersion()
-               psmdb.Spec.UnsafeConf = database.Spec.Engine.Replicas == 1
+               // We want the UnsafeConf to be immutable, so we only set it if the
+               // CR doesn't exist yet.
+               if psmdbCRNotFound {
+                       psmdb.Spec.UnsafeConf = database.Spec.Engine.Replicas == 1
+               }
                psmdb.Spec.Pause = database.Spec.Paused

                if database.Spec.Engine.Version == "" {
```

oksana-grishchenko:
> the following diff would actually guarantee the immutability of the psmdb.Spec.UnsafeConf  without exposing anything to the user.

looks like it should. @gen1us2k can we give it a try?

recharte:
> > the following diff would actually guarantee the immutability of the psmdb.Spec.UnsafeConf  without exposing anything to the user.
> 
> looks like it should. @gen1us2k can we give it a try?

IMO it's worth a try because it seems simple and elegant and avoids extra changes in the FE/BE to properly set the AllowUnsafeConfiguration field.

api/v1alpha1/databasecluster_types.go
```diff
@@ -196,6 +196,8 @@ type Monitoring struct {
 type DatabaseClusterSpec struct {
 	// Paused is a flag to stop the cluster
 	Paused bool `json:"paused,omitempty"`
+	// AllowUnsafeConfiguration field used to ensure that the user can create configurations unfit for production use.
+	AllowUnsafeConfiguration bool `json:"allowUnsafeConfiguration,omitempty"`
```
oksana-grishchenko, line 200: just posting this comment to prevent an accidental merging without addressing https://github.com/percona/everest-operator/pull/57#issuecomment-1702912316

gen1us2k:
> The thing is that with this change you don't guarantee immutability you just pass on that responsibility to the operator's user, which for the full-stack everest will be either the FE or the BE. So who will guarantee that this field won't be changed by updates?

This is not about guaranteeing the immutability of the field. This is just to solve the problem when a cluster scales from a single node to 3 nodes or even more. 

>I don't think this is the best approach because it gives the user a false sense of security and control for something that we could handle ourselves in the operator without exposing it to the user.


It's also not about providing a sense of security. It's just to provide a user-friendly way of communicating unsafe configurations that are not recommended for production use. 

>The current implementation doesn't guarantee immutability even if the user doesn't change it explicitly.
e.g. if the users set

I agree but we need to communicate and understand this option across all upstream operators with the Cloud team to understand potential use cases. I believe that we will use it later even for clustered deployments in some cases. A while back we used unsafe configurations for all deployments and now we use it depending on nodes in the cluster.

Another point about changing from 3 to 1. Yep, that's an issue right now and we can improve it in the following PRs after having a better understanding of how unsafeConfigurations work.

I set this rule for unsafe configurations depending on a number of replicas a while back because of a request from Sergey that it prevents the operator from following best practices from the security standpoint.

>I haven't tested it but I think that the following diff would actually guarantee the immutability of the psmdb.Spec.UnsafeConf without exposing anything to the user.

We do not need to guarantee the immutability of this field because there might be cases when we need to move from unsafe to safe configuration. However, your proposed solution has the following drawbacks

1. Unnecessary API request to the k8s API which may lead to rate limits on the client-go side which may affect operator stability on a large scale (remember we need to support 100 clusters reconciliation and we can easily hit rate limits here)
2. This logic depends on the network which might be not stable taking into account packet loss, connectivity problems, and other network issues, however, the original solution is network-agnostic
3. It does not provide a piece of information for the end user about unsafe configurations and we try to be smart here. Using allowUnsafeConfiguration we can have an indicator on the FE side that the following configuration of a database cluster is not recommended for production usage. It also prevents us from having integration with advisors (which might happen in the near future because all Percona products should be integrated with each other)
4. It lacks do not repeat yourself principle because we need to have almost the same logic for every reconciliation loop
5. It also increased cognitive complexity and cyclomatic complexity of reconciliation. 
6. It does not provide any piece of information to the end user about insecure configuration

In a nutshell, we can improve it later, There might be cases when we need to set the unsafe configuration to true for a specific use case but my original solution is not bulletproof to human mistakes yet it does not leave room for network errors and is simple and stupid enough to go with. 


> looks like it should. @gen1us2k can we give it a try?

I mean I can implement it but I do not see value here and I described the drawbacks above.

I believe that allowUnsafeConfiguration might be useful in the future and we can easily remove it later.

@recharte @oksana-grishchenko

api/v1alpha1/databasecluster_types.go
gen1us2k: I won't merge unless we are on the same page.

recharte:
@gen1us2k you may move forward and merge this solution. Let's unblock this fix.

PR #58, opened 2023-09-01
EVERESST-250 Add labels to the restore CR

oksana-grishchenko:
EVEREST-250 Add labels to the restore CR

Needed to list restores properly


recharte (approved):
Thanks, I missed this.

michal-kralik (approved):

PR #68, opened 2023-09-07
EVEREST-392 fix PMM monitoring

recharte:
EVEREST-392

* The PMM ServerHost can't contain the protocol nor the port, we must pass only the hostname.
* PMM API key integration with cluster secrets:

  * In order to keep auto-generating the rest of the cluster secret's data in PXC we can't create a secret with just the PMM API key. We need to wait for the secret to be autogenerated and then update it with the PMM API key.

  * In PSMDB we can create the secret right away with just the PMM API key and the rest of the secret's data will still be auto-generated.

  * In PG we shall not use the separate secret as the users. This can be
created before the DB because the operator doesn't change this secret.

P.S. I didn't update the e2e tests, too tired for it right now and we need to get this into the release tomorrow but I did test it manually. I'll add the automated tests in a follow-up PR.

### DIFF
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -23,6 +23,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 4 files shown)


oksana-grishchenko (approved):

hors:
@gen1us2k @oksana-grishchenko could you please check e2e test

gen1us2k (approved):

PR #73, opened 2023-09-12
EVEREST-380: support IP without subnet

michal-kralik:
EVEREST-380

- IP source range did not support IP without a subnet
- This commit adds support for IPs by adding a /32 subnet automatically


controllers/databasecluster_controller.go
```diff
@@ -2844,6 +2844,20 @@ func (r *DatabaseClusterReconciler) updateObject(ctx context.Context, obj, oldOb
 	return r.Update(ctx, obj)
 }
 
+func (r *DatabaseClusterReconciler) parseIPSourceRanges(ranges []string) []string {
```
gen1us2k, line 2847: Can we have tests for this function?
Also, what about using https://pkg.go.dev/net#ParseIP and https://pkg.go.dev/net#ParseCIDR?

michal-kralik: Updated

oksana-grishchenko (approved):

gen1us2k (approved):

PR #75, opened 2023-09-14
EVEREST-356 Remove finalizer steps and added delete pvc finalizer

gen1us2k:


oksana-grishchenko (approved):

controllers/databasecluster_controller.go
```diff
@@ -1009,15 +1015,10 @@ func (r *DatabaseClusterReconciler) reconcilePXC(ctx context.Context, req ctrl.R
...
-			database.Spec.Paused = current.Spec.Pause
+			database.Spec.Paused = pxc.Spec.Pause
 		}
 	}
 
-	pxc := &pxcv1.PerconaXtraDBCluster{}
-	err = r.Get(ctx, types.NamespacedName{Name: database.Name, Namespace: database.Namespace}, pxc)
-	if err != nil && !k8serrors.IsNotFound(err) {
-		return errors.Wrapf(err, "failed to get pxc object %s", database.Name)
-	}
```
recharte, line 1020: Why was a similar code snippet added to the PG reconciliation and removed here?

gen1us2k: Because when we were implementing restores I added this request above and operator were making two similar requests to get one spec which makes no sense. Right now it makes the only one request

recharte: ohhh I missed that, since it didn't show up in the diff, thanks

recharte (approved):

PR #82, opened 2023-09-19
EVEREST-433 fix pg restart on backup request

recharte:
EVEREST-433
We need to keep the PVC-based repo in the list even if the user defines a cloud-based repo because the PG operator will restart the cluster if the only repo in the list is changed.

### DIFF
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -1403,7 +1403,8 @@ func reconcilePGBackRestRepos(
 	}
 
 	availableRepoNames := []string{
-		"repo1",
+		// repo1 is reserved for the PVC-based repo, see below for details on
+		// how it is handled
 		"repo2",
 		"repo3",
 		"repo4",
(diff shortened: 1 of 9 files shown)


gen1us2k (approved):

michal-kralik (commented):
The description sounds like an issue in the PG operator.
PG shall not be restarted when backups are changed, right?

Wouldn't it make more sense to fix this in the PG operator instead?

oksana-grishchenko (approved):
Thanks for adding the descriptive comments! I think it's a good trade-off to handle the PG operator limitations

recharte:
> The description sounds like an issue in the PG operator. PG shall not be restarted when backups are changed, right?
> 
> Wouldn't it make more sense to fix this in the PG operator instead?

Actually, given the PGO architecture and even if it's very inconvenient for us I believe that the cluster needs to be restarted after such a change. I talked with Ege and Slava about this and they feel this way as well.

PR #83, opened 2023-09-19
EVEREST-200 delete SSL secrets

recharte:
EVEREST-200

In the same way that PR #75 added the delete PVC finalizer that ensures the deletion of PVCs as well as the DB connection secrets we shall also add the delete SSL finalizer to delete the auto-generated SSL certs secrets.

We don't need any finalizer for PSMDB because the SSL secrets have an owner reference to the PerconaServerMongoDB CR so they get automatically deleted when the DB cluster is deleted.

### DIFF
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -130,6 +130,8 @@ timeout server 28800s
 	finalizerDeletePXCPVC           = "delete-pxc-pvc"
 	finalizerDeletePSMDBPVC         = "delete-psmdb-pvc"
 	finalizerDeletePGPVC            = "percona.com/delete-pvc"
+	finalizerDeletePXCSSL           = "delete-ssl"
+	finalizerDeletePGSSL            = "percona.com/delete-ssl"
 )
 
 var operatorDeployment = map[everestv1alpha1.EngineType]string{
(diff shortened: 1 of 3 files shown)


controllers/databasecluster_controller.go
```diff
@@ -130,6 +130,8 @@ timeout server 28800s
 	finalizerDeletePXCPVC           = "delete-pxc-pvc"
 	finalizerDeletePSMDBPVC         = "delete-psmdb-pvc"
 	finalizerDeletePGPVC            = "percona.com/delete-pvc"
+	finalizerDeletePXCSSL           = "delete-ssl"
```
gen1us2k, line 133: What about psmdb?

recharte: I actually meant to put this PR as a draft, I'm still looking at PSMDB, sorry for that.

recharte: We don't need any finalizer for PSMDB because the SSL secrets have an owner reference to the `PerconaServerMongoDB` CR so they get automatically deleted when the DB cluster is deleted.

michal-kralik (approved):

oksana-grishchenko (approved):

PR #98, opened 2023-09-25
EVEREST-107 Upgraded to controller-runtime 0.16.2

gen1us2k:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-197

Controller-runtime package introduced some backward incompatible changes and cannot be upgraded via dependabot

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**


main.go
```diff
@@ -76,13 +78,18 @@ func main() {
...
-		MetricsBindAddress:     metricsAddr,
-		Port:                   9443,
+		Scheme: scheme,
+		Metrics: metricsserver.Options{
+			BindAddress: metricsAddr,
+		},
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "9094838c.percona.com",
-		Namespace:              ns,
```
oksana-grishchenko, line 85: The `Port` and `Namespace` fields where deleted in the new version, will the manager work correctly without them?

michal-kralik (approved):

main.go
gen1us2k: Tests are passing. Namespace moved to Cache.Namespaces and then renamed to Cache.DefaultNamespaces. Port is required for webhook settings and since we do not use them we do not need this. As for metrics it looks clear.

oksana-grishchenko (approved):

PR #99, opened 2023-09-26
EVEREST-115: azure storage

michal-kralik:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-115

**Related pull requests**

- https://github.com/percona/percona-everest-backend/pull/216

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**


api/v1alpha1/backupstorage_types.go
```diff
@@ -22,22 +22,26 @@ import (
...
-	// Type is a type of backup storage. Currently only S3 is supported.
-	// +kubebuilder:validation:Enum=s3
+	// Type is a type of backup storage.
+	// +kubebuilder:validation:Enum=s3;gcs;azure
 	Type BackupStorageType `json:"type"`
 	// Bucket is a name of bucket.
 	Bucket string `json:"bucket"`
 	// Region is a region where the bucket is located.
-	Region string `json:"region"`
+	Region string `json:"region,omitempty"`
```
gen1us2k, line 42: Starting from now we need to validate these fields on the BE side to be sure that provided configuration is valid and notify the end user if something is missing

controllers/databaseclusterrestore_controller.go
```diff
@@ -254,6 +254,12 @@ func (r *DatabaseClusterRestoreReconciler) restorePSMDB(ctx context.Context, res
 					Region:            backupStorage.Spec.Region,
 					EndpointURL:       backupStorage.Spec.EndpointURL,
 				}
+			case everestv1alpha1.BackupStorageTypeAzure:
+				psmdbCR.Spec.BackupSource.Azure = &psmdbv1.BackupStorageAzureSpec{
+					Container:         backupStorage.Spec.Bucket,
+					Prefix:            "everest",
```
gen1us2k, line 260: Can we make it constant somewhere? It appears multiple times hardcoded.

e2e-tests/tests/features/dbbackup_pg/20-assert.yaml
```diff
@@ -10,6 +10,22 @@ spec:
   backupStorageName: test-storage
   dbClusterName: test-pg-cluster
 ---
+apiVersion: everest.percona.com/v1alpha1
+kind: DatabaseClusterBackup
+metadata:
+  name: test-db-backup-azure
+spec:
+  backupStorageName: test-storage-azure
+  dbClusterName: test-pg-cluster
```
gen1us2k, line 19: Why do we need to have these asserts for S3 storage? Why are they created for PG cluster?

api/v1alpha1/backupstorage_types.go
michal-kralik: I agree

e2e-tests/tests/features/dbbackup_pg/20-assert.yaml
michal-kralik: What do you mean?
This PR is related to GCS and Azure. No changes to S3 have been done.

gen1us2k: Nevermind, I thought you made changes for asserts only.

.github/workflows/ci.yml
```diff
@@ -118,19 +118,19 @@ jobs:
...
-          echo "$BASE64_KUBECONFIG_DBAAS"|base64 -d > "${KUBECONFIG}"
-          kubectl kuttl test --config ./e2e-tests/kuttl-eks-core.yml
-        env:
-          PXC_OPERATOR_VERSION: 1.13.0
-          PSMDB_OPERATOR_VERSION: 1.14.0
-          PG_OPERATOR_VERSION: 2.2.0
-          BASE64_KUBECONFIG_DBAAS: ${{ secrets.BASE64_KUBECONFIG_DBAAS }}
-          PERCONA_VERSION_SERVICE_URL: https://check-dev.percona.com/versions/v1
+      # - name: Run core tests
+      #   run: |
```
gen1us2k, line 122: Uncomment it please

api/v1alpha1/backupstorage_types.go
gen1us2k: Where can I find the validation PR for that?

controllers/databasecluster_controller.go
```diff
@@ -328,14 +329,23 @@ func (r *DatabaseClusterReconciler) genPSMDBBackupSpec(
...
+				Type: psmdbv1.BackupStorageS3,
 				S3: psmdbv1.BackupStorageS3Spec{
 					Bucket:            backupStorage.Spec.Bucket,
 					CredentialsSecret: backupStorage.Spec.CredentialsSecretName,
 					Region:            backupStorage.Spec.Region,
 					EndpointURL:       backupStorage.Spec.EndpointURL,
 				},
 			}
+		case everestv1alpha1.BackupStorageTypeAzure:
+			storages[backup.Spec.BackupStorageName] = psmdbv1.BackupStorageSpec{
```
gen1us2k, line 341: what about gcs support for psmdb?

controllers/databasecluster_controller.go
```diff
@@ -906,6 +934,11 @@ func (r *DatabaseClusterReconciler) genPXCBackupSpec(
 				Region:            backupStorage.Spec.Region,
 				EndpointURL:       backupStorage.Spec.EndpointURL,
 			}
+		case everestv1alpha1.BackupStorageTypeAzure:
+			storages[backup.Spec.BackupStorageName].Azure = &pxcv1.BackupStorageAzureSpec{
+				ContainerPath:     backupStorage.Spec.Bucket,
```
gen1us2k, line 939: What about gcs support for PXC?

controllers/databasecluster_controller.go
```diff
@@ -1855,7 +2015,7 @@ func (r *DatabaseClusterReconciler) reconcilePGBackupsSpec(
 	return newBackups, nil
 }
 
-func (r *DatabaseClusterReconciler) genPGDataSourceSpec(ctx context.Context, database *everestv1alpha1.DatabaseCluster) (*crunchyv1beta1.DataSource, error) {
+func (r *DatabaseClusterReconciler) genPGDataSourceSpec(ctx context.Context, database *everestv1alpha1.DatabaseCluster) (*crunchyv1beta1.DataSource, error) { //nolint:gocognit,lll,maintidx
```
gen1us2k, line 2018: Can we split this function into smaller chunks instead of adding nolint for cognitive complexity, long line, and maintenance index?

controllers/databaseclusterrestore_controller.go
```diff
@@ -254,6 +253,12 @@ func (r *DatabaseClusterRestoreReconciler) restorePSMDB(ctx context.Context, res
 					Region:            backupStorage.Spec.Region,
 					EndpointURL:       backupStorage.Spec.EndpointURL,
 				}
+			case everestv1alpha1.BackupStorageTypeAzure:
```
gen1us2k, line 256: What about the support of GCS?

controllers/databaseclusterrestore_controller.go
```diff
@@ -313,6 +317,11 @@ func (r *DatabaseClusterRestoreReconciler) restorePXC(ctx context.Context, resto
...
```
gen1us2k, line 320: What about support of GCS?

e2e-tests/tests/features/dbbackup_pg/20-assert.yaml
```diff
@@ -10,6 +10,22 @@ spec:
...
+---
+apiVersion: everest.percona.com/v1alpha1
+kind: DatabaseClusterBackup
+metadata:
```
gen1us2k, line 23: The PR misses tests for PSMDB/PXC engines

e2e-tests/tests/features/dbbackup_pg/20-dbb.yaml
```diff
@@ -29,3 +29,55 @@ spec:
...
+type: Opaque
+stringData:
+  GCS_KEY: gcs_key
+---
+apiVersion: everest.percona.com/v1alpha1
+kind: BackupStorage
+metadata:
+  name: test-storage-gcs
+spec:
+  type: gcs
```
gen1us2k, line 73: IMO we should split these tests to independent ones

config/crd/bases/everest.percona.com_backupstorages.yaml
```diff
@@ -49,16 +49,15 @@ spec:
...
                 enum:
                 - s3
+                - gcs
+                - azure
                 type: string
             required:
             - bucket
             - credentialsSecretName
-            - endpointURL
-            - region
```
oksana-grishchenko, line 61: We're gonna need an update for the validation process on the BE side for that change - check that for the S3 the fields are set. Also we're gonna need to validate the access and write permissions for the GCS and Azure by create/update a backup storage. I couldn't find a ticket for it in Jira, could you please create one so it would be prominent there's work to be done to complete the feature?

oksana-grishchenko (commented):
That's a great new feature for Everest!

However it's not clear for me how the upstream operators are gonna know about GCS storage configuration, could you please re-request the review once it's done or elaborate how it's gonna work if I'm missing something.

controllers/databasecluster_controller.go
michal-kralik: Per our discussion, psmdb does not have GCS support yet.

controllers/databasecluster_controller.go
michal-kralik: Per our discussion, PXC does not have GCS support yet.

controllers/databaseclusterrestore_controller.go
michal-kralik: Per our discussion, psmdb does not have GCS support yet.

controllers/databaseclusterrestore_controller.go
michal-kralik: Per our discussion, PXC does not have GCS support yet.

e2e-tests/tests/features/dbbackup_pg/20-dbb.yaml
michal-kralik: Running the backups against a single cluster at the same time allows us complete the tests faster and discover concurrency issues, if there are any.

controllers/databasecluster_controller.go
michal-kralik: I observed we usually add `nolint` directives when it comes complexity or length of functions.
Other methods in the operator use `nolint` as well so I understand the linters serve more as a recommendation than a practice.
Do you think this practice is something we shall address in a more rigorous fashion?

config/crd/bases/everest.percona.com_backupstorages.yaml
michal-kralik: I agree. I'll look into into backend next. I wanted to make sure the approach in operator is fine first.

michal-kralik:
> However it's not clear for me how the upstream operators are gonna know about GCS storage configuration, could you please re-request the review once it's done or elaborate how it's gonna work if I'm missing something.

@oksana-grishchenko per our discussion, support for GCS is in the PG operator only at the moment.
Once PXC and PSMDB operators add support for GCS, we shall extend the implementation to cover these as well.

config/crd/bases/everest.percona.com_backupstorages.yaml
gen1us2k: The approach is fine, yet it would be best to merge it after the backend changes will be there because the spec needs to be updated.

gen1us2k:
>@oksana-grishchenko per our discussion, support for GCS is in the PG operator only at the moment.
Once PXC and PSMDB operators add support for GCS, we shall extend the implementation to cover these as well.

@michal-kralik What support are you talking about? Native one? That's true that the PG operator supports GCS at the moment, however, it's not true that PXC and PSMDB operators do not support GCS. They support it via S3 integration and they do not support the native one because they have no customers that want this.

config/crd/bases/everest.percona.com_backupstorages.yaml
michal-kralik: Wouldn't it make more sense to switch the order? i.e. merge this, take the schema and put it into backend?

If we were to merge backend first, the schema used in the backend would not be valid before the operator PR is merged.
Also the backend depends on the operator. The operator does not have to worry about the version of the backend.

This seems to me like a more rational approach. Am I missing anything?

controllers/databasecluster_controller.go
gen1us2k: We can move some parts of this function to different functions to make the logical split of the function

controllers/databaseclusterrestore_controller.go
gen1us2k: The conversation is not resolved. We need to add PXC/PSMDB handles for GCS storage and make the configuration using S3 fields for the PXC/PSMDB operators because S3 in those operators stands for GCS/S3 and these operators supports that. Also, We need to have tests for that as well

michal-kralik:
> @michal-kralik What support are you talking about? Native one? That's true that the PG operator supports GCS at the moment, however, it's not true that PXC and PSMDB operators do not support GCS. They support it via S3 integration and they do not support the native one because they have no customers that want this.

I meant native support. You can of course use GCS via S3. We support this option.

controllers/databaseclusterrestore_controller.go
gen1us2k: Yet, there should be a comment that explains this logic because it looks a bit misleading

gen1us2k:
>I meant native support. You can of course use GCS via S3. We support this option.

How? There are no cases for the GCS storage type. For the everest users, we should make this configuration clear without the confusion that we have for the upstream operators.

config/crd/bases/everest.percona.com_backupstorages.yaml
gen1us2k: The merging order does not make much sense but we need to have these updates first. Those changes are related and the PRs should be linked with each other using the 'Related PRs section'. Also, the validation should be present because right now we do not have any validation neither on the operator side nor on the BE side. The validation is also part of the implementation which is missing right now.

controllers/databasecluster_controller.go
michal-kralik: Absolutely.
If you feel we shall approach this in a more systematic way, let's open a new ticket and address this in a separate PR.

Since this approach is consistent with the rest of the codebase and does not make the codebase worse, I think it's fine and we can address the removal of selected `nolint` rules separately, if needed.

michal-kralik:
> How? There are no cases for the GCS storage type. For the everest users, we should make this configuration clear without the confusion that we have for the upstream operators.

You would follow the same approach as other tools employ for working with S3-compatible storages.
You would use the `type: s3` for backup storage, and properly configure `endpointURL` and optionally `region`.

This way you can talk to Google Storage, DigitalOcean, OpenStack Object Storage or any other S3-compatible storage.
It's a common pattern used across many different tools.
I'll be happy to explain this in more detail if it would be helpful.

gen1us2k:
>You would follow the same approach as other tools employ for working with S3-compatible storages.
You would use the type: s3 for backup storage, and properly configure endpointURL and optionally region.

This way you can talk to Google Storage, DigitalOcean, OpenStack Object Storage or any other S3-compatible storage.
It's a common pattern used across many different tools.
I'll be happy to explain this in more detail if it would be helpful.

How we will explain this to the end user since we have the GCS storage in our enum? I'm thinking more from the user perspective for our API users (FE team, CLI, and the rest users). As a guy that familiar with clouds I understand this point but our users don't (You can test this hypothesis on Peter or Danil and for them this experience would be frustrating and confusing).

gen1us2k (changes requested):
We need to review validation logic on the BE side to have the implementation between BE and operator side. Also https://github.com/percona/everest-operator/pull/99/files#r1339762781 and https://github.com/percona/everest-operator/pull/99/files#r1339764451 needs to be resolved is provided implementation and not using just a silent click on the Resolve conversation button

gen1us2k:
>This way you can talk to Google Storage, DigitalOcean, OpenStack Object Storage or any other S3-compatible storage.
It's a common pattern used across many different tools.
I'll be happy to explain this in more detail if it would be helpful.

This also makes sense but we have its own logic for the PG operator and let's keep it consistent across all upstream operators to avoid confusion for the end users. Right now we need to tell users that the end user needs to select gcs for PG only and s3 for PXC/PSMDB. Also, we use S3 SDK on the BE side and in that case, without those changes, we need to check this additionally because I'm not sure that it will work for the gcs using S3 SDK.

As for other storage locations we do not have this in our roadmap and it's the topic of conversation for Q1 or even later

michal-kralik:
> How we will explain this to the end user since we have the GCS storage in our enum? I'm thinking more from the user perspective for our API users (FE team, CLI, and the rest users). As a guy that familiar with clouds I understand this point but our users don't (You can test this hypothesis on Peter or Danil and for them this experience would be frustrating and confusing).

With the backend validation you won't be able to use GCS type with other than a PG database.
You will receive an error message which may feature a link to the documentation. 
A link to documentation can also be part of the frontend during the process of the backup storage configuration.
I think there's plenty of opportunities for us to make this a pleasant experience.

The docs can explain how to use GCS via S3 with PXC/PSMDB since it's very straightforward. It's a matter of setting the url to `https://storage.googleapis.com`

We may even link to a Google Cloud, DigitalOcean, OpenStack documentation where this is documented.
Or even other S3-compatible services' documentation since there's plenty of them available and it does not make sense to have a separate `type` for each of these.

controllers/databasecluster_controller.go
hors: @gen1us2k in this case, please fix it as well :) we need to have a general rule for our whole code

https://github.com/percona/percona-everest-backend/pull/158/files#diff-4485a00dd63af8089d0eda57e3745c25f0da2e1d2efe9f9038a2d93f9bf9559fR150
https://github.com/percona/percona-everest-backend/pull/158/files#diff-fa107c8ea97a323473ac245ac9066b063fde00bc9c445d3b1740303dda5901b2R140

gen1us2k: @hors https://jira.percona.com/browse/EVEREST-482 on it. Agree

gen1us2k:
>With the backend validation you won't be able to use GCS type with other than a PG database.
You will receive an error message which may feature a link to the documentation.

Here comes the confusion for the API users. I agree that's a matter of setting the URL yet since we have the GSC type enabled for PG the best option for me is to have this for all operators to keep everything consistent

recharte:
Hey @michal-kralik @gen1us2k I haven't reviewed the PR itself but I was looking at this discussion regarding the GCS support and I want to share my opinion.

I understand the reluctance to add a case for GCS in PXC/PSMDB as these operators don't have native GCS support. However, if we can ask users for a GCS type configuration and then convert it internally to a supported type (S3) then I think we should do this.

A related example:
I see similarities with the way the secrets for the backup storage credentials are handled. For PXC/PSMDB we simply take the credentials secret in the format we ask the user (with the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` key-values) but for PG we need to take this generic secret format and convert it to a format that can be understood by PGBackRest.
We could ask the user to use a different secret format just for PG because that's needed by the operator but that's a bad UX. Instead, to facilitate the UX, we take a generic one and convert it when needed.
Similarly, I think we can take a GCS backup storage type and convert it to an S3 one if a given operator doesn't natively support it. This will definitely make for a far better UX that asking the users to handle it differently for each operator. 

If we have some technical limitation that doesn't allow us to do this conversion internally then maybe we should skip the GCS type entirely and explain to the users how to convert GCS to S3 to be able to use it for any operator. Having GCS only being usable by PG doesn't seem great.

michal-kralik:
Support for GCS can be generally added in three ways:

1. PG will have a native S3, Azure and GCS support. 
PXC/PSMDB will have a native S3 and Azure support, no support for `type: gcs`.
2. PG/PXC/PSMDB will have a native S3 and Azure support and all will have a `type: gcs` option which internally uses S3.
3. PG will have a native S3, Azure and GCS support.
PXC/PSMDB will have a native S3/Azure support and will have a `type: gcs` option which internally uses S3.

The option 1 requires the users to understand GCS can be used with an S3-compatible clients. This can be easily explained in the UI by providing guides how to generate credentials. It's is also a common knowledge various object storages provide a common S3 interface.

The option 2 will make it challenging for us to migrate to a native GCS support in the future because the S3 credentials are not compatible with GCS native access. Since we would already have a `type: gcs` we may need to introduce a new option, such as `type: gcs-native` or make the API backwards incompatible.

The option 3 requires a change to our backend API so that backup storage supports multiple sets of credentials. This is because GCS native uses a different set of credentials than S3-compatible GCS access and you cannot convert between one and the other. It's not possible to reuse the same credentials for GCS access in PG and PXC/PSMDB.

Considering all of the above, from my experience, the option 1 provides the best experience and is well understood by users as it's common approach across other solutions as well.

If you guys feel otherwise, let me know which option you agreed on and I'll be happy to rework the implementation.

michal-kralik:
Documenting our decision - @recharte confirmed we shall drop GCS support to resolve this.

michal-kralik:
@recharte this is ready for review now

oksana-grishchenko (approved):

gen1us2k (approved):

gen1us2k:
Thanks. Looks good but validation improvements are still required

controllers/databasecluster_controller.go
```diff
@@ -430,14 +449,23 @@ func (r *DatabaseClusterReconciler) genPSMDBBackupSpec(
...
 					EndpointURL:       backupStorage.Spec.EndpointURL,
 				},
 			}
+		case everestv1alpha1.BackupStorageTypeAzure:
+			storages[schedule.BackupStorageName] = psmdbv1.BackupStorageSpec{
+				Type: psmdbv1.BackupStorageAzure,
+				Azure: psmdbv1.BackupStorageAzureSpec{
+					Container:         backupStorage.Spec.Bucket,
+					Prefix:            azureStoragePrefix,
+					CredentialsSecret: backupStorage.Spec.CredentialsSecretName,
```
gen1us2k, line 466: Setting resources is required now for backup pods
https://github.com/percona/everest-operator/commit/2636ed4e5e4979bf26cd5b9cc97e86f84b2e042e

gen1us2k (changes requested):

controllers/databasecluster_controller.go
michal-kralik: I don't think I understand.
They're set in https://github.com/percona/everest-operator/blob/be76a10d78c5ba6c2dab90c5fc839d3835b9320b/controllers/databasecluster_controller.go#L303-L308

Are there any other resource limits to be configured?

gen1us2k: They are set by default for psmdb, but for PXC the configuration is set per storage
https://github.com/percona/everest-operator/blob/be76a10d78c5ba6c2dab90c5fc839d3835b9320b/controllers/databasecluster_controller.go#L938-L943

gen1us2k: as for PG they set by default as well and there's no need to add any changes

michal-kralik: ok, so for PG no action.
What would be your expectation for PXC?
We already set the resource limits in https://github.com/percona/everest-operator/blob/be76a10d78c5ba6c2dab90c5fc839d3835b9320b/controllers/databasecluster_controller.go#L938-L944

This applies to both S3 and Azure storages. Am I missing anything here?

gen1us2k: Yeah. That works. Thanks

gen1us2k (approved):

PR #106, opened 2023-10-05
EVEREST-470 Set default resources for backup/proxy pods

gen1us2k:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-470

PXC backup container uses resources, but we do not set them . We need to set resources for backup as well.

**Solution:**

Since @hors reported the issue and had a discussion with him about this as the first solution. The operator adopted the default resources provided by operators in cr.yaml file. Yet, we should revisit this once we start working on templates

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -1909,6 +1909,7 @@ func (r *DatabaseClusterReconciler) genPGDataSourceSpec(ctx context.Context, dat
 				"--type=immediate",
 				"--set=" + backupBaseName,
 			},
+			// XXX: Remove this once templates will be available
 			Resources: corev1.ResourceRequirements{
 				Limits: corev1.ResourceList{
 					corev1.ResourceMemory: resource.MustParse("128Mi"),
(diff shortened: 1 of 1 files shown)


controllers/databasecluster_controller.go
```diff
@@ -1894,6 +1909,12 @@ func (r *DatabaseClusterReconciler) genPGDataSourceSpec(ctx context.Context, dat
 				"--type=immediate",
 				"--set=" + backupBaseName,
 			},
+			Resources: corev1.ResourceRequirements{
+				Limits: corev1.ResourceList{
+					corev1.ResourceMemory: resource.MustParse("128Mi"),
+					corev1.ResourceCPU:    resource.MustParse("200m"),
+				},
+			},
```
oksana-grishchenko, line 1917: will this part stay when the templates available? if not, let's have a comment here also about removing it. And for the limits too if relevant.

oksana-grishchenko (approved):

PR #112, opened 2023-10-09
EVEREST-524 Removed custom configuration for HAProxy

gen1us2k:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-524

The operator uses the wrong format of the HAProxy config. It's required to pass the whole HAProxy config for the configuration field. You can see the example in the [documentation](https://docs.percona.com/percona-operator-for-mysql/pxc/haproxy-conf.html#passing-custom-configuration-options-to-haproxy)

Also, PXC operator have these parameters as a [new default](https://github.com/percona/percona-docker/commit/deb1e6f03d42cccf4d2e15f1880a1d5a13d0c194) so these options are not needed anymore

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -107,10 +107,6 @@ wsrep_trx_fragment_size=3670016
 	pxcMinimalConfigurationTemplate = `[mysqld]
 wsrep_provider_options="gcache.size=%s"
 `
-	haProxyDefaultConfigurationTemplate = `timeout client 28800s
-timeout connect 100500
-timeout server 28800s
-`
 	psmdbDefaultConfigurationTemplate = `
       operationProfiling:
         mode: slowOp
@@ -1046,9 +1042,6 @@ func (r *DatabaseClusterReconciler) reconcilePXC(ctx context.Context, req ctrl.R
 
 	if database.Spec.Proxy.Type == "" {
 		database.Spec.Proxy.Type = everestv1alpha1.ProxyTypeHAProxy
-	}
-	if database.Spec.Proxy.Type == everestv1alpha1.ProxyTypeHAProxy && database.Spec.Proxy.Config == "" {
-		database.Spec.Proxy.Config = haProxyDefaultConfigurationTemplate
 	}
 	if err := r.Update(ctx, database); err != nil {
 		return err


controllers/databasecluster_controller.go
```diff
@@ -106,10 +106,6 @@ wsrep_trx_fragment_size=3670016
 `
 	pxcMinimalConfigurationTemplate = `[mysqld]
 wsrep_provider_options="gcache.size=%s"
-`
-	haProxyDefaultConfigurationTemplate = `timeout client 28800s
-timeout connect 100500
-timeout server 28800s
```
oksana-grishchenko, line 112: A bit concerned we just delete the default configuration. If the format is incorrect as the ticket says, is it possible to make it correct?

controllers/databasecluster_controller.go
```diff
@@ -1047,9 +1043,6 @@ func (r *DatabaseClusterReconciler) reconcilePXC(ctx context.Context, req ctrl.R
 	if database.Spec.Proxy.Type == "" {
 		database.Spec.Proxy.Type = everestv1alpha1.ProxyTypeHAProxy
 	}
-	if database.Spec.Proxy.Type == everestv1alpha1.ProxyTypeHAProxy && database.Spec.Proxy.Config == "" {
-		database.Spec.Proxy.Config = haProxyDefaultConfigurationTemplate
-	}
```
oksana-grishchenko, line 1052: Is the config required for the upstream operators to work correctly? Am I understand it correctly that If we don't set the Proxy config here then some default values from the upstream operators will be applied?

controllers/databasecluster_controller.go
gen1us2k: The format is incorrect and we need to specify the whole HAProxy config (see the attached link to the documentation in the description of PR). Also, since we're changing timeouts it makes no sense because it's defaults that are set in the PXC operator (see the second link to the actual change in the PXC operator)

controllers/databasecluster_controller.go
oksana-grishchenko: ok, I see the description. resolving the question

controllers/databasecluster_controller.go
oksana-grishchenko: I believe we don't need it anymore according to the description, nevernid

oksana-grishchenko (approved):

controllers/databasecluster_controller.go
gen1us2k: Yep. That's right the default configuration will be applied, however, we need to provide this option on the UI and in our documentation about changing defaults for proxies. The config is not required per se because all upstream operators have their own default configs.

PR #116, opened 2023-10-10
EVEREST-342 Added default configuration for PXC

gen1us2k:
**Better default configuration for created PXC clusters **
---
**Problem:**
EVEREST-342

PXC wasn't using the correct configuration.

**Solution:**
Generated configuration for mysqld was generated using mysqlcalculator and added as an option for small/medium/large dimensions of a database cluster in Everest. The solution lacks dynamic config generation but it will be implemented in the upcoming releases after the actual integration with mysqlcalculator

The configuration might look a bit weird but we do not have an agreement about the dimensions of how we should treat the requested compute resources. Right now we treat it as the limits for a database engine pod without taking into account additional resources such as backup pods, proxy pods, and pmm pods.

Marco's calculator treats the compute request as a request for the whole node deployment (pxc, haproxy, and pmm) and calculates requests/limits automatically as well as liveness problems. It needs to be agreed on the product side first before the actual integration.

Also, the solution lacks tests but since it will be reworked in the near future, before the actual integration with the calculator tests will be added. 

**CHECKLIST**
---
**Jira**

**Tests**


controllers/databasecluster_controller.go
```diff
@@ -129,13 +121,20 @@ wsrep_provider_options="gcache.size=%s"
...
+		everestv1alpha1.DatabaseEnginePXC:        pxcDeploymentName,
+		everestv1alpha1.DatabaseEnginePSMDB:      psmdbDeploymentName,
+		everestv1alpha1.DatabaseEnginePostgresql: pgDeploymentName,
+	}
 
-var maxUnavailable = intstr.FromInt(1)
+	maxUnavailable       = intstr.FromInt(1)
+	exposeAnnotationsMap = map[ClusterType]map[string]string{
+		ClusterTypeEKS: {
+			"service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout": "4000",
```
recharte, line 134: I remember we removed the other LB annotations when we moved from PMM's DBaaS to everest, is this annotation required? Does it work without the other annotations we had previously?

controllers/databasecluster_controller.go
```diff
@@ -1043,6 +1050,17 @@ func (r *DatabaseClusterReconciler) reconcilePXC(ctx context.Context, req ctrl.R
...
+	if database.Spec.Engine.Config == "" {
+		if database.Spec.Engine.Resources.Memory.CmpInt64(memorySmallSize) >= 0 && database.Spec.Engine.Resources.Memory.CmpInt64(memoryMediumSize) < 0 {
+			database.Spec.Engine.Config = pxcConfigSizeSmall
+		}
+		if database.Spec.Engine.Resources.Memory.CmpInt64(memoryMediumSize) >= 0 && database.Spec.Engine.Resources.Memory.CmpInt64(memoryLargeSize) < 0 {
+			database.Spec.Engine.Config = pxcConfigSizeMedium
+		}
+		if database.Spec.Engine.Resources.Memory.CmpInt64(memoryLargeSize) >= 0 {
+			database.Spec.Engine.Config = pxcConfigSizeLarge
+		}
```
recharte, line 1062: What if it's smaller than `memorySmallSize`?

gen1us2k: Made it work for dimensions sent from FE

controllers/databasecluster_controller.go
gen1us2k: It does work without annotations

controllers/databasecluster_controller.go
```diff
@@ -1043,6 +1049,17 @@ func (r *DatabaseClusterReconciler) reconcilePXC(ctx context.Context, req ctrl.R
...
+	if database.Spec.Engine.Config == "" {
+		if database.Spec.Engine.Resources.Memory.Cmp(memorySmallSize) == 0 {
+			database.Spec.Engine.Config = pxcConfigSizeSmall
+		}
+		if database.Spec.Engine.Resources.Memory.Cmp(memoryMediumSize) == 0 {
+			database.Spec.Engine.Config = pxcConfigSizeMedium
+		}
+		if database.Spec.Engine.Resources.Memory.Cmp(memoryLargeSize) == 0 {
+			database.Spec.Engine.Config = pxcConfigSizeLarge
+		}
```
oksana-grishchenko, line 1061: Didn't get it. The `Cmp` returns 0 if the quantity is equal to the value. What if the memory is something in between? There were that intervals before, why it's not the case now?

gen1us2k: Frontend supports only 3 dimensions. The configuration might not work if memory has changed.

oksana-grishchenko (approved):

PR #115, opened 2023-10-10
EVEREST-532 Support for PSMDB operator 1.15

gen1us2k:
**PSMDB 1.15 support**
---
**Problem:**
EVEREST-532

**CHECKLIST**
---
**Jira**
- [x] Is the Jira ticket created and referenced properly?

**Tests**
- [ ] Is an Integration test/test case added for the new feature/change?
- [ ] Are unit tests added where appropriate?

PR #133, opened 2023-10-24
EVEREST-532 fix mongo scheduled backups

recharte:
**Fix mongo scheduled backups**
---
**Problem:**
EVEREST-532

MongoDB scheduled backups didn't show the cluster name and thus weren't being properly reconciled. 

**Related pull requests**

- #115

**Cause:**
In psmdb operator v1.15.0 the PSMDBCluster field is no longer being filled.

**Solution:**
Move to the new ClusterName field instead of PSMDBCluster.

**CHECKLIST**
---
**Jira**

**Tests**
- ~[ ] Is an Integration test/test case added for the new feature/change?~
- ~[ ] Are unit tests added where appropriate?~

### DIFF
--- a/deploy/bundle.yaml
+++ b/deploy/bundle.yaml
@@ -56,15 +56,14 @@ spec:
                 description: Region is a region where the bucket is located.
                 type: string
               type:
-                description: Type is a type of backup storage. Currently only S3 is supported.
+                description: Type is a type of backup storage.
                 enum:
                 - s3
+                - azure
                 type: string
             required:
             - bucket
             - credentialsSecretName
-            - endpointURL
-            - region
             - type
             type: object
           status:
--- a/bundle/manifests/everest-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/everest-operator.clusterserviceversion.yaml
@@ -103,7 +103,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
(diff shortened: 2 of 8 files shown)


deploy/bundle.yaml
```diff
@@ -56,15 +56,14 @@ spec:
...
+                description: Type is a type of backup storage.
                 enum:
                 - s3
+                - azure
                 type: string
             required:
             - bucket
             - credentialsSecretName
-            - endpointURL
-            - region
```
oksana-grishchenko, line 67: I assume this is a leftover from the Azure support changes, the bundle wasn't rebuild back then.

michal-kralik (approved):

oksana-grishchenko (approved):
So this is a new version of the operator. Should we reference it in catalog, BE and cli for the 0.4.0 RC?

deploy/bundle.yaml
recharte: Yes that's it.

recharte:
> So this is a new version of the operator. Should we reference it in catalog, BE and cli for the 0.4.0 RC?

If there wasn't an update to the bundle we could just rebuild the operator v0.4.0 image and nothing else would need to change. With the change to the bundle I believe we'll also need to rebuild the catalog v0.4.0 image but I think that's it.

PR #143, opened 2023-11-03
EVEREST-587: wait for db before restore

michal-kralik:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-587

We did not wait for DB before initiating a restore.

**Cause:**
DB was not ready and the restore failed.

**Solution:**
Wait for DB before restore

**CHECKLIST**
---
**Jira**

**Tests**


oksana-grishchenko (approved):
PG uses a different mechanism, I believe it's not affected by the problem?

michal-kralik:
> PG uses a different mechanism, I believe it's not affected by the problem?

Yes, PG seems to handle this differently and is thus not affected.

gen1us2k (approved):

recharte (approved):

PR #155, opened 2023-11-09
EVEREST-606: fix for unlimited PG backup retention

michal-kralik:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-606

**Cause:**
By default we have RetentionCopies=0 which means infinity for PXC and PSMDB but for the PG 

 
To have an equivalent behavior in PG we should set repo-retention-full=9999999 the maximum allowed

**Solution:**
Properly set unlimited retention

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -1566,8 +1566,13 @@ func reconcilePGBackRestRepos(
 			// Keep track of backup storages which are already in use by a repo
 			backupStoragesInRepos[backupSchedule.BackupStorageName] = struct{}{}
 
+			copies := int(backupSchedule.RetentionCopies)
+			if copies == 0 {
+				// Zero copies means unlimited. PGBackRest supports values 1-9999999
+				copies = 9999999
+			}
 			pgBackRestGlobal[fmt.Sprintf(pgBackRestPathTmpl, repo.Name)] = "/"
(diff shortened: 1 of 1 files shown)


controllers/databasecluster_controller.go
```diff
@@ -1566,8 +1566,13 @@ func reconcilePGBackRestRepos(
 			// Keep track of backup storages which are already in use by a repo
 			backupStoragesInRepos[backupSchedule.BackupStorageName] = struct{}{}
 
+			copies := int(backupSchedule.RetentionCopies)
+			if copies == 0 {
+				// Zero copies means unlimited. PGBackRest supports values 1-9999999
+				copies = 9999999
+			}
```
oksana-grishchenko, line 1573: to avoid using the same magic constant twice we may have it in a separate function

oksana-grishchenko (approved):
Thanks for catching and fixing this!

controllers/databasecluster_controller.go
michal-kralik: We could, but given we ignore code reuse all over operator, I just decided to follow the same pattern :)
Fixed now

recharte (approved):

recharte:
> We could add this field to the asserts in the e2e-tests.

Apparently we're missing e2e-tests for the PG scheduled backups altogether :sweat: I'm pretty sure we had them before, don't know what happened to them...

michal-kralik:
> We could add this field to the asserts in the e2e-tests.

I also thought about this but figured the full config is a string and I'm not sure if our e2e tests can check only for a substring there.

PR #157, opened 2023-11-13
EVEREST-490 Ensure CRVersion is immutable during the upgrade

gen1us2k:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-490

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**


e2e-tests/tests/upgrade/pxc/30-assert.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: kuttl.dev/v1
+kind: TestAssert
+timeout: 100
+---
+apiVersion: pxc.percona.com/v1
+kind: PerconaXtraDBCluster
+metadata:
+  name: test-pxc-cluster
+spec:
+  crVersion: 1.12.0
```
oksana-grishchenko, line 10: can we somehow get the `crVersion` here and in the other tests as the `PXC_OPERATOR_VERSION`/`PREVIOUS_PXC_OPERATOR_VERSION` env vars? otherwise we would still need to update all the tests when we update the values in the config which makes the whole idea of having them as variables less beneficial.

oksana-grishchenko (approved):
Thanks for the improvement! Approving to unblock the PR, please consider the comments

e2e-tests/tests/upgrade/pxc/30-assert.yaml
gen1us2k: Unfortunately no.

gen1us2k: Kuttl does not support it

PR #162, opened 2023-11-15
EVEREST-518 PITR api

oksana-grishchenko:
**Point-in-time recovery API**
---
**Problem:**
EVEREST-518 PITR API

**CHECKLIST**
---
**Jira**

**Tests**


gen1us2k (approved):

api/v1alpha1/databasecluster_types.go
```diff
@@ -225,6 +228,26 @@ type Backup struct {
...
+	// PITR is the configuration of the point in time recovery
+	PITR PITRSpec `json:"pitr,omitempty"`
+}
+
+// PITRSpec represents a specification to configure point in time recovery for a database backup/restore.
+type PITRSpec struct {
+	// Enabled is a flag to enable PITR
+	Enabled bool `json:"enabled,omitempty"`
+	// StorageName is the name of the BackupStorage where the PITR is enabled
+	StorageName string `json:"storageName"`
```
recharte, line 240: For consistency, maybe we should call this `BackupStorageName`, what do you think?

api/v1alpha1/databasecluster_types.go
```diff
@@ -225,6 +228,26 @@ type Backup struct {
...
+	// TimeBetweenUploadsSec number of seconds between the binlogs uploads
+	TimeBetweenUploadsSec *int `json:"timeBetweenUploadsSec,omitempty"`
```
recharte, line 242: What about `UploadIntervalSec`?

api/v1alpha1/databasecluster_types.go
```diff
@@ -225,6 +228,26 @@ type Backup struct {
...
+	// CompressionType type of compression (psmdb only, default s2)
+	// +kubebuilder:validation:Enum:=gzip;snappy;lz4;s2;pgzip;zstd
+	CompressionType compress.CompressionType `json:"compressionType,omitempty"`
+	// CompressionLevel level of compression (psmdb only, more info
+	// https://docs.percona.com/percona-backup-mongodb/reference/backup-options.html#backupcompressionlevel)
+	CompressionLevel *int `json:"compressionLevel,omitempty"`
+	// Resources the compute resource requirements (pxc only, default no limits applied)
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
```
recharte, line 250: Do we really need to expose these to the users right now or could we just set some sensible defaults ourselves in the controller?
We've been bitten by a complex API in the past which eventually lead to a big refactor a few months ago to remove a lot of options that most users didn't care about. These seem similar to the type of things we removed back then, I don't want us to make the same mistake of making the interface too bloated.

api/v1alpha1/databaseclusterrestore_types.go
```diff
@@ -37,6 +40,53 @@ type DatabaseClusterRestoreStatus struct {
...
+// PITR represents a specification to configure point in time recovery for a database backup/restore.
+type PITR struct {
+	// Type is the type of recovery. pxc: full list; psmdb, pg: date,latest
+	// +kubebuilder:validation:Enum:=date;latest;transaction;skip
+	// +kubebuilder:default:=date
+	Type string `json:"type,omitempty"`
+	// Date is the date to recover to
+	Date *RestoreDate `json:"date,omitempty"`
+	// GTID is the transaction ID to restore to (pxc only)
+	GTID string `json:"gtid,omitempty"`
```
recharte, line 52: Transaction level selection isn't part of the scope for the first iteration of PITR we'll only have the date and time option. Maybe we should leave this out of the picture for now.

oksana-grishchenko: Sure, I love deleting things :) Done

api/v1alpha1/databasecluster_types.go
oksana-grishchenko: Slava mentioned once he would add them, but I don't mind deleting it. Let's start with smth simple.
Done.

api/v1alpha1/databasecluster_types.go
oksana-grishchenko: Agree, updated

api/v1alpha1/databasecluster_types.go
oksana-grishchenko: Good point, updated, thx

recharte (approved):

PR #169, opened 2023-11-16
EVEREST-496 Multi namespace support

gen1us2k:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-496

That's a small PoC for cluster-wide operators that stores backup storage (the same applies for monitoring configs) in the percona-everest and replicates secrets across namespaces that it manages.

Pros:
1. BE needs to manage only one BackupStorage or MonitoringConfig with the referenced secret in only one namespace
2. Operator handles the state of the secret and updates it across all namespaces where it was created 

Cons:
1. Increased blast radius. When (not if, but when) we will release a broken version of the operator DBaaS features will be broken for all database clusters across all namespaces. Having it in the namespace mode locks it only for the one namespace and this approach is recommended by Percona. Read more https://www.percona.com/blog/percona-operators-deployment-explained-delving-into-cluster-wide-vs-namespace-scoped/
2. Bloated service account permissions to run everest operator in the cluster-wide mode. Instead of having one SA per namespace with the least privileges. If percona-everest namespace is compromised, the attacker can get access to all namespaces that manage everest-operator 
3. Additional load to the QA team because they need to test two different modes of everest operator and this is an additional step for their workflow 
4. The configuration becomes more fragile because CLI needs to set env var for `percona-everest` namespace for two components: BE and operator
5. During the update of a secret and when the operator fails to reconcile it in any namespace the end user will stay unnotified without any messaging from the everest side
6. There is no separation of concerns for the sensitive data because everything is stored in the one namespace yet I believe we need to store it in different namespaces to follow security best practices 

Yet we have an additional request for the cluster-wide operator to use admission webhooks to move validation logic from the BE side to the operator, however, the scope of change is huge because 

1. Would be best to implement multi-namespace mode for the operator without additional logic, ensure that everything works as expected
2. Ensure that everest operator can work with the different versions of the upstream operators
3. Implement reconcile loops for monitoring configs, and backup storages
4. Implement admission webhoooks or any other stuff

**CHECKLIST**
---
**Jira**

**Tests**


config/manager/manager.yaml
```diff
@@ -103,7 +103,7 @@ spec:
             cpu: 10m
             memory: 64Mi
         env:
-        - name: WATCH_NAMESPACE
+        - name: EVEREST_NAMESPACE
```
gen1us2k, line 106: This namespace will be primary, yet during the deployment we need to set WATCH_NAMESPACES=namespace1,namespace2 to run reconciliation logic there

config/manifests/bases/everest-operator.clusterserviceversion.yaml
```diff
@@ -56,9 +56,9 @@ spec:
     type: OwnNamespace
   - supported: true
     type: SingleNamespace
-  - supported: false
-    type: MultiNamespace
   - supported: true
```
gen1us2k, line 59: Yet I have no idea of how it will work with the RBAC stuff

controllers/databasecluster_controller.go
```diff
@@ -141,7 +141,8 @@ var (
 // DatabaseClusterReconciler reconciles a DatabaseCluster object.
 type DatabaseClusterReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme           *runtime.Scheme
+	everestNamespace string
```
gen1us2k, line 145: Storing the value of EVEREST_NAMESPACE env var

controllers/databasecluster_controller.go
```diff
@@ -329,6 +330,12 @@ func (r *DatabaseClusterReconciler) genPSMDBBackupSpec(
 		if err != nil {
 			return psmdbv1.BackupSpec{Enabled: false}, errors.Join(err, fmt.Errorf("failed to get backup storage %s", backup.Spec.BackupStorageName))
 		}
+		if database.Namespace != r.everestNamespace {
+			if err := r.reconcileSecret(ctx, backupStorage, database); err != nil {
```
gen1us2k, line 334: Create or update secret in the targetNamespace

controllers/databasecluster_controller.go
```diff
@@ -3234,3 +3244,27 @@ func (r *DatabaseClusterReconciler) defaultPSMDBSpec() *psmdbv1.PerconaServerMon
...
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+	err = r.Get(ctx, types.NamespacedName{Name: backupStorage.Spec.CredentialsSecretName, Namespace: r.everestNamespace}, secret)
+	if err != nil {
+		return err
+	}
+	secret.Namespace = database.Namespace
+	if !backupStorage.IsNamespaceAllowed(database.Namespace) {
+		return fmt.Errorf("%s namespace is not allowed to use for %s backup storage", database.Namespace, backupStorage.Name)
```
gen1us2k, line 3259: I assume we need to improve validation on the BE side as well

main.go
```diff
@@ -106,14 +122,15 @@ func main() {
...
 		os.Exit(1)
 	}
-
-	if err = (&controllers.DatabaseClusterReconciler{
+	dbReconciler := &controllers.DatabaseClusterReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
+	}
+	if err := dbReconciler.SetupWithManager(mgr); err != nil {
```
gen1us2k, line 129: Yep, I know that the controller for BackupStorage is missing to run the reconciliation loop either for backupstorage update or for a secret update

api/v1alpha1/backupstorage_types.go
```diff
@@ -74,3 +78,21 @@ type BackupStorageList struct {
 func init() {
 	SchemeBuilder.Register(&BackupStorage{}, &BackupStorageList{})
 }
+func (b *BackupStorage) UpdateNamespacesList(namespace string) {
+	if b.Status.Namespaces == nil {
+		b.Status.Namespaces = make(map[string]bool)
+	}
+	b.Status.Namespaces[namespace] = true
```
oksana-grishchenko, line 85: we only add the namespaces to the Status but not delete them, it means we may end up in the situation when a namespace is listed in the Status but not even exists in the cluster. Is it an intended trade-off?

api/v1alpha1/backupstorage_types.go
```diff
@@ -74,3 +78,21 @@ type BackupStorageList struct {
...
+}
+func (b *BackupStorage) IsNamespaceAllowed(namespace string) bool {
```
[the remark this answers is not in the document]
oksana-grishchenko: What should happen with an existing db cluster if the SRE removes it's namespace from the targetNamespaces?

api/v1alpha1/backupstorage_types.go
```diff
@@ -45,10 +45,14 @@ type BackupStorageSpec struct {
 	Description string `json:"description,omitempty"`
 	// CredentialsSecretName is the name of the secret with credentials.
 	CredentialsSecretName string `json:"credentialsSecretName"`
+	// TargetNamespaces is the list of namespaces where the operator will copy secrets provided in the CredentialsSecretsName.
+	TargetNamespaces []string `json:"targetNamespaces"`
```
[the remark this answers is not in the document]
oksana-grishchenko: We'll also need to validate that the listed namespaces exist

oksana-grishchenko (commented):
In general the idea with `targetNamespaces` for the allowed list and `status.namespaces` for the actual list looks good, however to keep the lists consistent we would need to handle the deletion of 
- the targetNamespaces
- the actual namespaces in the cluster

api/v1alpha1/backupstorage_types.go
```diff
@@ -45,10 +45,14 @@ type BackupStorageSpec struct {
...
 }
 
 // BackupStorageStatus defines the observed state of BackupStorage.
-type BackupStorageStatus struct{}
+type BackupStorageStatus struct {
+	Namespaces map[string]bool `json:"usedNamespaces"`
```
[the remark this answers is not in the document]
recharte: I don't think we need this, the idea would be to use the same logic we use for [creating the PMM secret](https://github.com/percona/everest-operator/blob/14c108b7a9b92c43f65fd2b800f0f31bbc029689/controllers/databasecluster_controller.go#L652-L656) or the [backup secret for PG](https://github.com/percona/everest-operator/blob/14c108b7a9b92c43f65fd2b800f0f31bbc029689/controllers/databasecluster_controller.go#L1975-L1985).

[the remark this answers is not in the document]
gen1us2k: We need this because I don't want to violate security best practices

api/v1alpha1/backupstorage_types.go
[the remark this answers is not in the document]
gen1us2k: It will be deleted

controllers/databasecluster_controller.go
```diff
@@ -2572,18 +2618,7 @@ func (r *DatabaseClusterReconciler) addPXCKnownTypes(scheme *runtime.Scheme) err
 }
 
 func (r *DatabaseClusterReconciler) addPSMDBKnownTypes(scheme *runtime.Scheme) error {
-	version, err := r.getOperatorVersion(context.Background(), types.NamespacedName{
```
gen1us2k, line 2575: We can drop support for the 1.12 PSMDB operator

api/v1alpha1/backupstorage_types.go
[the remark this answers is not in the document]
oksana-grishchenko: acceptable

oksana-grishchenko (approved):
The implementation itself looks good, my concerns are resolved. Approving the PR to unblock it in case the massive "Cons" section is considered by the product as the limitations we can live with.

recharte (commented):
Overall the PR looks like it is functionally correct however, I think the followed approach is overly complex.

I argue that we don't need the BackupStorage controller at all. We already have an [index](https://github.com/percona/everest-operator/blob/fdaa0adc929a918419f14418fa36709771bb85c6/controllers/databasecluster_controller.go#L2677-L2695) and [watcher](https://github.com/percona/everest-operator/blob/fdaa0adc929a918419f14418fa36709771bb85c6/controllers/databasecluster_controller.go#L2780-L2785) configured in the DBC controller that runs the reconciliation loop every time a secret referenced by a BS triggers an event.
Therefore, we can use this same mechanism to create or update the end secret without needing to keep track of any namespace. In a sense we can think of this secret as being linked to a specific DBC than to the generic BS.

Moreover, we already use this approach for PG because the secret format needs to be different than the one specified in the BS which is the same for PXC/PSMDB. See [here](https://github.com/percona/everest-operator/blob/fdaa0adc929a918419f14418fa36709771bb85c6/controllers/databasecluster_controller.go#L1995-L2005) and [here](https://github.com/percona/everest-operator/blob/fdaa0adc929a918419f14418fa36709771bb85c6/controllers/databasecluster_controller.go#L1336-L1369)

gen1us2k:
> I argue that we don't need the BackupStorage controller at all. We already have an [index](https://github.com/percona/everest-operator/blob/fdaa0adc929a918419f14418fa36709771bb85c6/controllers/databasecluster_controller.go#L2677-L2695) and [watcher](https://github.com/percona/everest-operator/blob/fdaa0adc929a918419f14418fa36709771bb85c6/controllers/databasecluster_controller.go#L2780-L2785) configured in the DBC controller that runs the reconciliation loop every time a secret referenced by a BS triggers an event.
Therefore, we can use this same mechanism to create or update the end secret without needing to keep track of any namespace. In a sense we can think of this secret as being linked to a specific DBC than to the generic BS.


> Moreover, we already use this approach for PG because the secret format needs to be different than the one specified in the BS which is the same for PXC/PSMDB. See [here](https://github.com/percona/everest-operator/blob/fdaa0adc929a918419f14418fa36709771bb85c6/controllers/databasecluster_controller.go#L1995-L2005) and [here](https://github.com/percona/everest-operator/blob/fdaa0adc929a918419f14418fa36709771bb85c6/controllers/databasecluster_controller.go#L1336-L1369)


Well, this was my first approach to solving the issue. This works fine for the backupStorages or monitoringConfigs in the single namespace scenario. 

Since the idea is to create backupstorages or monitoring configs in the default namespace (percona-everest) the existence of the objects is only in the default namespace. However, databasecluster controller replicates secrets across namespaces (since they are namespaced) there's no way (at least I know) to bind that secret with the backupStorage or monitoringConfig. Since the logic for backup storage and monitoring config is the same I'll describe the flow based on backup storage

Let's take a look at the backup storage In the default namespace. To ensure that the secret will be deleted we can use ownerreferences here and we don't quite need a controller here. The backend or some other controller can create it for you yet, once the backupstorage is deleted, the secret will be deleted as well.

Since secrets are namespaced to enable the multi namespace support in the cluster-wide mode we need to replicate them to the targetNamespace during the reconciliation of database clusters(because this is the only place when we need to ensure that the secret exists). We can't use ownerreferences here because they are not designed for that and we need to come up with a way to store the information where the secret was created. Here are the labels to link the secret in the different namespace with the existing backup storage. 

After testing this using indexers and watchers, when the user tries to delete the secret from a different namespace the reconciliation loop does not start because there's no event generated/passed to the operator after the deletion and it ends in the inconsistent state.

Hence, we need to have an additional controller to ensure that after backup storage deletion, all secrets across all namespaces will be created and here we have finalizers as an additional handle to block the deletion. We need to delete all secrets across all namespaces except the default one (because we use owner references here and the deletion might block here because of cross-linking between entities)

Since the backupstorage controller is namespaced (because we create backup storages only in the default namespace) we need to have a way to watch for secrets and run reconciliation of the backup storage entity on the secret change. For the default secret in the default namespace we have `Owns` for that and this works fine. For all other namespaces, we need to listen to secrets and run a reconcile loop only for labeled secrets.

I hope that it explains the solution after a couple of days of trying to implement as simply as I can.

recharte (changes requested):
This is such an important change that we can't merge it without proper tests for it.

controllers/databasecluster_controller.go
```diff
@@ -3258,3 +3295,77 @@ func (r *DatabaseClusterReconciler) defaultPSMDBSpec() *psmdbv1.PerconaServerMon
...
+	backupStorage *everestv1alpha1.BackupStorage,
+	database *everestv1alpha1.DatabaseCluster,
+) error {
+	secret := &corev1.Secret{}
+	err := r.Get(ctx, types.NamespacedName{Name: backupStorage.Spec.CredentialsSecretName, Namespace: database.Namespace}, secret)
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+	}
```
recharte, line 3310:
```suggestion
	if err != nil && !k8serrors.IsNotFound(err) {
		return err
	}
```

controllers/databasecluster_controller.go
```diff
@@ -3258,3 +3295,77 @@ func (r *DatabaseClusterReconciler) defaultPSMDBSpec() *psmdbv1.PerconaServerMon
...
+		},
+	}
+	if !backupStorage.IsNamespaceAllowed(database.Namespace) {
+		return fmt.Errorf("%s namespace is not allowed to use for %s backup storage", database.Namespace, backupStorage.Name)
+	}
+	err = r.Create(ctx, secret)
+	if err != nil {
+		return err
+	}
+	return nil
```
recharte, line 3333:
```suggestion
	return r.Create(ctx, secret)
```

controllers/databasecluster_controller.go
```diff
@@ -3258,3 +3295,77 @@ func (r *DatabaseClusterReconciler) defaultPSMDBSpec() *psmdbv1.PerconaServerMon
...
+	monitoringConfig *everestv1alpha1.MonitoringConfig,
+	database *everestv1alpha1.DatabaseCluster,
+) error {
+	secret := &corev1.Secret{}
+	err := r.Get(ctx, types.NamespacedName{Name: monitoringConfig.Spec.CredentialsSecretName, Namespace: database.Namespace}, secret)
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+	}
```
recharte, line 3347:
```suggestion
	if err != nil && !k8serrors.IsNotFound(err) {
		return err
	}
```

controllers/databasecluster_controller.go
```diff
@@ -3258,3 +3295,77 @@ func (r *DatabaseClusterReconciler) defaultPSMDBSpec() *psmdbv1.PerconaServerMon
...
+		},
+	}
+	if !monitoringConfig.IsNamespaceAllowed(database.Namespace) {
+		return fmt.Errorf("%s namespace is not allowed to use for %s backup storage", database.Namespace, monitoringConfig.Name)
+	}
+	err = r.Create(ctx, secret)
+	if err != nil {
+		return err
+	}
+	return nil
```
recharte, line 3370:
```suggestion
	return r.Create(ctx, secret)
```

recharte (commented):
While reviewing I found a couple of tiny code style improvements, these are just some nitpicks so feel free to ignore them.

api/v1alpha1/backupstorage_types.go
```diff
@@ -45,10 +45,14 @@ type BackupStorageSpec struct {
 	Description string `json:"description,omitempty"`
 	// CredentialsSecretName is the name of the secret with credentials.
 	CredentialsSecretName string `json:"credentialsSecretName"`
+	// TargetNamespaces is the list of namespaces where the operator will copy secrets provided in the CredentialsSecretsName.
+	TargetNamespaces []string `json:"targetNamespaces,omitempty"`
```
recharte, line 49: This needs to be discussed with Peter and Nuna, giving the user this responsibility wasn't part of the product requirements.

recharte: Agreed on this approach.

e2e-tests/tests/features/multinamespace_pg/10-assert.yaml
```diff
@@ -0,0 +1,97 @@
...
+kind: BackupStorage
+metadata:
+  name: test-storage-scheduled
+spec:
+  type: s3
+  bucket: test-backup-bucket-scheduled
+  credentialsSecretName: test-storage-scheduled
+  region: us-east-2
+  endpointURL: s3.amazonaws.com
+---
```
recharte, line 42:
```suggestion
```

recharte (approved):
Thank you for adding the tests :hugs:.
I think that the only important case we're missing would be to validate the TargetNamespaces feature works as intended but that's part of the error cases which we aren't testing anywhere so I can't say this is mandatory for now.

PR #173, opened 2023-11-17
EVEREST-236: multiple db backups in a single bucket

michal-kralik:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-236

Multiple DBs cannot be properly stored in the same backup storage location.

**Solution:**
- Add support for multiple db clusters using the same backup storage location
- Retrieve PSMDB backup destination from PSMDB CR

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -104,7 +104,6 @@ const (
 	backupStorageNameField          = ".spec.backup.schedules.backupStorageName"
 	credentialsSecretNameField      = ".spec.credentialsSecretName" //nolint:gosec
 
-	azureStoragePrefix              = "everest"
 	databaseClusterNameLabel        = "clusterName"
 	monitoringConfigNameLabel       = "monitoringConfigName"
 	backupStorageNameLabelTmpl      = "backupStorage-%s"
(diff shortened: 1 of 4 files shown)


controllers/databasecluster_controller.go
```diff
@@ -3101,6 +3121,10 @@ func getPGRetentionCopies(retentionCopies int32) int {
 	return copies
 }
 
+func backupStoragePrefix(db *everestv1alpha1.DatabaseCluster) string {
+	return fmt.Sprintf("%s/%s", db.Name, db.UID)
+}
```
oksana-grishchenko, line 3126: the case when we have two BS referencing the same bucket is still possible, so do we need the storage name there?

michal-kralik: I don't think we do.
Backup storage defines a way to access a certain storage. In case two backup storages point to the same location, it's similar to having two different sets of credentials.

michal-kralik: Are you suggesting we shall have a check in place to make sure no two backup storages refer to the same location (e.g. the same bucket)?

oksana-grishchenko: That could be an option. Another option is have the storage name in the prefix. 
Both are handling the case with the same bucket so it won't break the functionality.

gen1us2k (approved):
Thanks for the implementation. It looks good so far yet Oksana's concerns should be addressed before merging

PR #204, opened 2023-12-06
EVEREST-603 PITR PXC Restore

oksana-grishchenko:
**PITR PXC restore**
---
**Problem:**
EVEREST-608

- PITR PXC Restore
- fixes for to the date format
- test for restore CR creation

**CHECKLIST**
---
**Jira**

**Tests**


api/v1alpha1/databasecluster_types.go
```diff
@@ -284,6 +284,8 @@ type DatabaseClusterStatus struct {
 	Size int32 `json:"size,omitempty"`
 	// Message is extra information about the cluster
 	Message string `json:"message,omitempty"`
+	// PITRUploadIntervalSec is the interval set for PITR to upload the binlogs.
+	PITRUploadIntervalSec int `json:"pitrUploadIntervalSec,omitempty"`
```
oksana-grishchenko, line 288: We could avoid having this new field however. Since the FE doesn't have an option to edit the uploadInterval, the interval is not going to be more than 10 min (the maximum from PSMDB) and we can just hardcode those 10 min to the FE.

recharte: What's the intended use case for this field? Is it for the FE to calculate the most recent usable PIT?

recharte: I just read the PR description
> Reason: for the first iterations it was decided to use heuristic approach to figure out the earliest and the latest restorable date. Earliest is the date of the latest backup and latest is now minus the upload interval.

recharte: Which everest component will be responsible for calculating the available time range using that heuristic?

oksana-grishchenko: The FE

recharte: For public reference, me and @oksana-grishchenko just had a call to discuss this, here is the summary.
In the future, we'll forcibly need to give the FE the available time ranges, so we might as well implement the interface for it right away.
Having said this, the idea is that the BE implements an endpoint that provides this time range to the FE. In this first version, the BE will implement the aforementioned heuristic and can use harcoded values for the upload interval so there's no need to have this field in the DBC CR's status.

oksana-grishchenko: UPD: 
Discussed the approach. To prevent the changes from the FE side when the actual intervals are gonna be available (and other information like `gtid`s) it was decided to remove this new field from the Operator and create an additional BE endpoint like `databaseclusters/{id}/pitr` which will provide the earliest and the latest dates

recharte (approved):

michal-kralik (approved):

PR #217, opened 2023-12-14
EVEREST-682: copy credentials during db provisioning

michal-kralik:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-682

- In case a DB is to be provisioned from a data source, we copy the credentials from the original DB
- This ensures the DB can be properly provisioned

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -208,6 +208,13 @@ func (r *DatabaseClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 	}
 
+	if database.Spec.DataSource != nil && database.Spec.DataSource.DBClusterBackupName != "" {
+		// We don't handle database.Spec.DataSource.BackupSource in operator
+		if err := r.copyCredentialsFromDBBackup(ctx, database.Spec.DataSource.DBClusterBackupName, database); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
 	if database.Spec.Engine.Type == everestv1alpha1.DatabaseEnginePXC {
 		err := r.reconcilePXC(ctx, req, database)
 		return reconcile.Result{}, err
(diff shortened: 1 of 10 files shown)


gen1us2k (approved):

e2e-tests/tests/features/dbbackup_pg/25-assert.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: test-pg-source
+status:
+  status: ready
+---
+apiVersion: pgv2.percona.com/v2
+kind: PerconaPGCluster
+metadata:
+  name: test-pg-source
+status:
+  state: ready
```
oksana-grishchenko, line 17: should we check if the new secret is there as well?

controllers/databasecluster_controller.go
```diff
@@ -2141,18 +2203,48 @@ func (r *DatabaseClusterReconciler) genPGDataSourceSpec(ctx context.Context, dat
...
+			dest = strings.TrimPrefix(dest, "s3://")
+		case everestv1alpha1.BackupStorageTypeAzure:
+			dest = strings.TrimPrefix(dest, "azure://")
+		}
+
+		dest = strings.TrimPrefix(dest, backupStorage.Spec.Bucket)
+		dest = strings.TrimLeft(dest, "/")
+		prefix := backupStoragePrefix(database)
+		prefixCount := len(strings.Split(prefix, "/"))
+		dest = "/" + strings.Join(strings.SplitN(dest, "/", prefixCount+1)[0:prefixCount], "/")
```
oksana-grishchenko, line 2240: It would be great if we could cover this transformation logic with a unit test that would also document the expected output

oksana-grishchenko (approved):

e2e-tests/tests/features/dbbackup_pg/25-assert.yaml
michal-kralik: I'm not sure how we would check for this.
The new secret is going to be there - either created by us or the operator.
We would need to compare it's the same as the other one, but I don't think kuttl supports this type of check.

oksana-grishchenko: Makes sense, ok

controllers/databasecluster_controller.go
oksana-grishchenko: Awesome, thanks!

e2e-tests/tests/features/dbbackup_pxc/81-delete-restore.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1
+kind: TestStep
+timeout: 10
+commands:
+  - command: kubectl -n "${NAMESPACE} dbr --all"
+  - command: kubectl -n "${NAMESPACE} pxc-restore --all"
```
oksana-grishchenko, line 6: a bit confused, what is the purpose of this commands?

michal-kralik: it's a bug. It should have included `delete`

PR #226, opened 2023-12-19
EVEREST-708 DB restoration status improvements

oksana-grishchenko:
**DB restoration status improvements**
---
**Problem:**
EVEREST-708

The FE needs to know when the DB cluster is restoring so any actions over it could be disabled. 
After the BE/FE sync it was decided:

- use `AppStateRestoring` DB cluster status to indicate when the cluster is restoring
- keep the original upstream status strings to be able to provide additional information about the restoration process (like what exact step of the restoration is happening now) in the future


api/v1alpha1/databaseclusterrestore_types.go
```diff
@@ -144,3 +147,74 @@ type DatabaseClusterRestoreList struct {
...
+	case pxcv1.RestoreStartCluster:
+		return false
+	case pxcv1.RestorePITR:
+		return false
+	case pxcv1.RestoreFailed:
+		return true
+	case pxcv1.RestoreSucceeded:
+		return true
+	}
+	return true
```
michal-kralik, line 183: I suggest we turn around all three functions to return `false` by default and only for selected statuses we return `true`.
Otherwise if a new status is added and we forget to update these functions, the new status will be considered as a "complete" status.

michal-kralik (approved):

gen1us2k (approved):

api/v1alpha1/databaseclusterrestore_types.go
recharte: That's a tough one :sweat_smile:

I'm leaning to agree with you @oksana-grishchenko.

If they add an end state and we have default `false` then there's no coming back.
In the other hand, if they add a transitory state and we say it's complete (default `true`), despite being wrong it's a transitory thing that will fix itself shortly.

Both case are wrong and we should really try to be in sync with the cloud team whenever they change this kind of stuff (I think that we do a good job of it with our weekly DBaaS sync). In the case this slips through, I think we should go with the option that fixes itself automatically, i.e. keeping `true` as a default.

oksana-grishchenko: Both approaches have their pro and cons, but looks like we have a consensus regarding the topic. Got an approval from Michal to resolve the thread.

PR #243, opened 2024-01-10
EVEREST-751 PSMDB restore to a new cluster fix

oksana-grishchenko:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-0

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databaseclusterrestore_controller.go
+++ b/controllers/databaseclusterrestore_controller.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"strings"
 	"time"
 
 	pgv2 "github.com/percona/percona-postgresql-operator/pkg/apis/pgv2.percona.com/v2"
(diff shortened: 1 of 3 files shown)


controllers/databaseclusterrestore_controller.go
```diff
@@ -246,7 +246,7 @@ func (r *DatabaseClusterRestoreReconciler) restorePSMDB(
 			}
 
 			psmdbCR.Spec.BackupSource = &psmdbv1.PerconaServerMongoDBBackupStatus{
-				Destination: fmt.Sprintf("s3://%s/%s", backupStorage.Spec.Bucket, restore.Spec.DataSource.BackupSource.Path),
+				Destination: restore.Spec.DataSource.BackupSource.Path,
```
recharte, line 249: Does this change affect the "normal" restores to the same DB?

recharte: Since this is only for the case when the backup source is used I guess not because we currently only use the backup name.

recharte (approved):

controllers/databaseclusterrestore_controller.go
recharte: P.S. feel free to resolve this conversation and merge the PR.

oksana-grishchenko: Restore to the same db works, just checked. You're right, we never actually used the backup source for the restores, that's why I believe we didn't notice the weird way of constructing the path

PR #244, opened 2024-01-11
EVEREST-520 Enable PSMDB PITR

oksana-grishchenko:
**Enable PSMDB PITR**
---
**Problem:**
EVEREST-520 

Enable psmdb pitr. 
Allow to use only one storage for PSMDB, update tests to handle that.

**Related PRs:**
- https://github.com/percona/percona-everest-backend/pull/386 

**CHECKLIST**
---
**Jira**

**Tests**


recharte (approved):

e2e-tests/tests/features/dbbackup_psmdb/20-assert.yaml
```diff
@@ -10,14 +10,6 @@ spec:
...
   dbClusterName: test-psmdb-cluster
 ---
-apiVersion: everest.percona.com/v1alpha1
-kind: DatabaseClusterBackup
-metadata:
-  name: test-db-backup-azure
-spec:
-  backupStorageName: test-storage-azure
-  dbClusterName: test-psmdb-cluster
----
```
oksana-grishchenko, line 20: two storages can not exist simultaneously from now on, so I moved the azure-related stuff to a separate step

michal-kralik: What is the reason we don't support multiple backup storage locations?

oksana-grishchenko: because PITR can not be enabled for PSMDB when multiple storages are used.

api/v1alpha1/databasecluster_types.go
```diff
@@ -284,6 +284,8 @@ type DatabaseClusterStatus struct {
 	Size int32 `json:"size,omitempty"`
 	// Message is extra information about the cluster
 	Message string `json:"message,omitempty"`
+	// DefaultStorage is the storage used by default (psmdb only)
+	DefaultStorage string `json:"defaultStorage,omitempty"`
```
recharte, line 288: Do we really need this? I really don't like it, it's a confusing field :sweat_smile:

e2e-tests/tests/features/dbbackup_psmdb/20-assert.yaml
recharte: it's a PBM limitation

api/v1alpha1/databasecluster_types.go
oksana-grishchenko: Well, me too, but how will the FE know what storage is a default one if we apply it deeply under the hood based on what is used for schedules/on-demand backups?

oksana-grishchenko: Another option would be to add to the BE the psmdb dependency so it could check the psmdb cluster directly, but I would rather avoid this

e2e-tests/tests/features/dbbackup_psmdb/20-assert.yaml
michal-kralik: Does this change prevent users from having multiple backup locations with different policies set per S3 bucket, such as expiration time?
If yes, I think we shall rethink the approach.

oksana-grishchenko: I believe the different policies mean the different buckets, which in terms of Everest means the different `BackupStorages`, so yes, it's gonna prevent users for having that. I understand the concern. 

It's again more like a product question - what is more important, the PITR or the multi storages support. Waiting for the product decision.

api/v1alpha1/databasecluster_types.go
recharte: So we had a couple of offline discussions and we'll move with this approach but we might rename this to `ActiveStorage`.

e2e-tests/tests/features/dbbackup_psmdb/20-assert.yaml
oksana-grishchenko: UPD: 
The [discussion](https://perconacorp.slack.com/archives/C0545J2BEJX/p1705491681958579) with Product folks and the Cloud team. 
**Conclusion:** move forward with the one storage limitation, created [the ticket](https://perconadev.atlassian.net/browse/PBM-1228) for PBM team to support multiple storages.

recharte (approved):

PR #265, opened 2024-02-04
EVEREST-633 Reconcile VMAgent and use a dedicated monitoring namespace

recharte:
**Reconcile VMAgent and use a dedicated monitoring namespace**
---
**Note for reviewers: Still missing some e2e-tests to assert the VMAgent reconciliation. While I work on these you can start to take a look at the code**
**Problem:**
EVEREST-633

By moving to a multi-namespace everest-operator installation we configured the operator group of the `percona-everest` namespace to have N target namespaces.
The VM Operator doesn't support multi-namespaces so we can't deploy it to the same namespace as the everest-operator.
In this PR we expect a separate namespace for managing everything monitoring related.
This monitoring namespace shall hold the MonitoringConfigs which will then get reconciled into a VMAgent CR which the VMOperator will use for pushing the metrics collected by the `kube-state-metrics`. This way, as soon as users create a MonitoringConfig, the k8s metrics will start flowing into the registered PMM instances. This removes the need of running `everestctl monitoring enable`.

Also, in this PR we remove the need for replicating the MonitoringConfig's secret into every namespace. The DBC controller can fetch the secret data directly from the monitoring namespace thus bypassing the need for complex secret replication.

**CHECKLIST**
---
**Jira**

**Tests**


main.go
```diff
@@ -46,8 +46,9 @@ import (
 )
 
 const (
-	defaultNamespaceEnvVar = "DEFAULT_NAMESPACE"
-	watchNamespacesEnvVar  = "WATCH_NAMESPACES"
+	systemNamespaceEnvVar     = "SYSTEM_NAMESPACE"
+	monitoringNamespaceEnvVar = "MONITORING_NAMESPACE"
+	dbNamespacesEnvVar        = "DB_NAMESPACES"
```
michal-kralik, line 51: Are these values mandatory? If yes, are we missing validation for them?

michal-kralik (approved):

main.go
recharte: Yes and no.
We have the validation [here](https://github.com/percona/everest-operator/blob/e5ad5afe3859f9bb3210058a89463bb24a15b790/main.go#L82-L97).
But I just found that the last one is missing an exit :see_no_evil: I'll fix it.

michal-kralik: This part validates if the environment variable exists. It doesn't check if it's not empty, does not have restricted characters, etc.

recharte: Ohh I see your point. We should check if the variable it's empty but we don't need to check for other characters or anything like that because it will fail to create the controller if the namespace doesn't exist and it will return an  appropriate message like this
```
[manager] 2024-02-06T13:54:40Z	ERROR	setup	unable to create controller	{"controller": "DatabaseCluster", "error": "namespaces \"awrongone\" not found"}
```

oksana-grishchenko (approved):
Looks solid. Haven't tried to actually run it and see if it's going to work and we're not missing anything important, but I hope you did :)  
Great job!

e2e-tests/tests/features/multinamespace_psmdb/10-assert.yaml
```diff
@@ -70,11 +71,3 @@ spec:
...
           region: us-east-2
         type: s3
-status:
-  replsets:
-    rs0:
-      initialized: true
-      ready: 1
-      size: 1
-      status: ready
-  state: ready
```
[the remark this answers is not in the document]
recharte: Well, I removed it because it was always timing out and we don't check it for the other DBEngines.
Giving it a second, I think we should though.
I'll add this when I add the VMAgent tests in a follow-up PR.
Thank you for pointing it out.

recharte:
> Looks solid. Haven't tried to actually run it and see if it's going to work and we're not missing anything important, but I hope you did :) Great job!

:sweat_smile: Yes, I definitely have been testing it and I'll continue testing it during the integration with the FE, I want this to be solid :sweat_smile:

PR #282, opened 2024-02-14
EVEREST-838 Don't delete scheduled backups from s3

recharte:
**Don't delete scheduled backups from s3**
---
**Problem:**
EVEREST-838

Scheduled backups were being deleted from s3.

**Cause:**
By default the pxc and psmdb operators set a finalizer for the scheduled backup.

**Solution:**
We don't want to delete these backups from s3 so we remove the finalizer.

**CHECKLIST**
---
**Jira**

**Tests**
- ~[ ] Is an Integration test/test case added for the new feature/change?~
- ~[ ] Are unit tests added where appropriate?~


oksana-grishchenko (approved):
Does PG already work as expected (does not delete data for both types of backups)?

PR #309, opened 2024-03-07
Fix MySQL PXC and HAProxy configs

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-824

There is an issue with the probe timeouts in PXC and HAProxy that leads to the system crashing because it takes too long to respond. 

**Solution:**
Fix the MySQL PXC and HAProxy configs.

**CHECKLIST**
---
**Jira**

**Tests**


mayankshah1607:
Review this commit for the mysqld config changes: https://github.com/percona/everest-operator/pull/309/commits/72253a8c15222d0d1b7800f57ff9dddc049b766f

mayankshah1607:
@oksana-grishchenko looks like I missed that part, sorry. I have added it to this PR now. :)

controllers/pxc_config.go
```diff
@@ -159,3 +164,41 @@ wsrep_trx_fragment_unit = bytes
...
+
+var ( //nolint:dupl
+	// A pxcResourceRequirementsSmall is the resource requirements for PXC for small clusters.
+	pxcResourceRequirementsSmall = corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("1.61Gi"),
+			corev1.ResourceCPU:    resource.MustParse("570m"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("1.6Gi"),
```
oksana-grishchenko, line 176: the limit is smaller then the request, I believe it should be 1.7Gi (1825361100 bytes = 1.699999999255Gi)

oksana-grishchenko (approved):
Overall looks good to me👍
A bit concerned that the resources/limits/timeouts are distributed through the code which makes it harder to maintain, but I can't see any easy solution how we could put it into a single config, so I believe at this point it's ok. 

I would also like Diogo to take a look at this PR since he is historically more into the topic, please do not merge until his review.

controllers/pxc_config.go
mayankshah1607: You're right, that's a typo. Thanks for catching it!

controllers/databasecluster_controller.go
```diff
@@ -968,6 +981,27 @@ func (r *DatabaseClusterReconciler) genPXCHAProxySpec(
...
 
 	haProxy.PodSpec.Configuration = database.Spec.Proxy.Config
+	if haProxy.PodSpec.Configuration == "" {
+		haProxy.PodSpec.Configuration = haProxyConfigDefault
+	}
+
+	// Ensure there is a env vars secret for HAProxy
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pxcHAProxyEnvSecretName,
```
recharte, line 991: Will we ever need to customize this per DB? Right now we are setting a global secret that will be used by every PXC DBC in a given namespace, which is fine because the value is hardcoded but will we need to customize this is the future?

controllers/databasecluster_controller.go
```diff
@@ -1444,12 +1495,15 @@ func (r *DatabaseClusterReconciler) reconcilePXC(ctx context.Context, req ctrl.R
...
 
-		if monitoring.Spec.Type == everestv1alpha1.PMMMonitoringType {
+		if monitoring.Spec.Type == everestv1alpha1.PMMMonitoringType { //nolint:nestif
 			image := defaultPMMClientImage
 			if monitoring.Spec.PMM.Image != "" {
 				image = monitoring.Spec.PMM.Image
 			}
 
+			//nolint:godox
+			// TODO (EVEREST-824): Set PMM container LivenessProbes and ReadinessProbes timeouts once possible.
```
recharte, line 1505: Why isn't it possible?

mayankshah1607: https://github.com/percona/percona-xtradb-cluster-operator/blob/c066c6ef6e440dd2a523d718c8f1851de866d3f3/pkg/apis/pxc/v1/pxc_types.go#L514

It doesn't expose an API for it unlike the rest :/ 
I think we would have to create a ticket for it so that it gets added..

controllers/databasecluster_controller.go
mayankshah1607: Hmm, good question. I am not sure about that.

If we want to customize this secret per DB, we might possibly also want to customize the HAProxy config file as well. Unless we want the user to configure these, I don't think we will have the need to customize this..

controllers/databasecluster_controller.go
recharte: Can you discuss this with Slava? If he agrees then please proceed to open a ticket for the appropriate projects in Jira [PXCO](https://perconadev.atlassian.net/browse/K8SPXC), [PSMDBO](https://perconadev.atlassian.net/browse/K8SPSMDB), [PGO](https://perconadev.atlassian.net/browse/K8SPG), [PSO](https://perconadev.atlassian.net/browse/K8SPS). Maybe Slava will tell you to open only for one of them :shrug: please confirm with him :slightly_smiling_face:

controllers/databasecluster_controller.go
recharte: Let's keep it like this for now, if we find the need to change it in the future we can do so

recharte (approved):
Good job :tada:

controllers/databasecluster_controller.go
recharte: If you decide to open a ticket then please update this comment with the ticket ID

mayankshah1607: Done

PR #317, opened 2024-03-13
Fix MySQL PITR restore to new cluster

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
[EVEREST-890](https://perconadev.atlassian.net/browse/EVEREST-890)

PITR restore of a MySQL to a new cluster fails.

**Cause:**
When performing a PITR for MySQL, the created pxc-restore object always assumes its own bucket as the backup source. This will not work when we want to PITR restore to a new cluster, as it should use the source cluster's bucket and not its own.

**Solution:**

Updated the PXC restore code so that it uses the same bucket as the one used for the source backup. This way it works for same-cluster and cross-cluster PITR restore.

**CHECKLIST**
---
**Jira**

**Tests**

[EVEREST-890]: https://perconadev.atlassian.net/browse/EVEREST-890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/controllers/databaseclusterrestore_controller.go
+++ b/controllers/databaseclusterrestore_controller.go
@@ -328,7 +328,8 @@ func (r *DatabaseClusterRestoreReconciler) restorePXC(
 			pxcCR.Spec.BackupName = restore.Spec.DataSource.DBClusterBackupName
 		}
 
-		if dataSource := restore.Spec.DataSource; dataSource.BackupSource != nil {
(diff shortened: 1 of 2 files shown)


recharte (approved):
Looks good :muscle:

PR #348, opened 2024-03-27
EVEREST-939 | Ensure Secrets for PITR BackupStorages are copied

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-939

When we have the PITR backupstroage different from the scheduled backups, the Secret from `everest-system` namespace is not copied.

**Solution:**
Ensure that Secrets are copied over similar to how we do for Scheduled backups.

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -1161,7 +1161,7 @@ func (r *DatabaseClusterReconciler) genPXCBackupSpec( //nolint:gocognit
 		}
 	}
 
-	if database.Spec.Backup.PITR.Enabled {
+	if database.Spec.Backup.PITR.Enabled { //nolint:nestif
 		storageName := *database.Spec.Backup.PITR.BackupStorageName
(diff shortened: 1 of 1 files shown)


oksana-grishchenko (approved):
Thanks Mayank for the fix! 

I tried to reproduce it from scratch with the 0.8.0 Operator - it's reproducible:
```
NAME                                               READY   STATUS                       RESTARTS   AGE
mysql-nry-haproxy-0                                2/2     Running                      0          2m20s
mysql-nry-pitr-66c46d7bdf-xfzxr                    0/1     CreateContainerConfigError   0          24s
mysql-nry-pxc-0                                    1/1     Running                      0          2m20s
```
and the backups got stuck also, so exactly the case described in the ticket.

For this branch the problem is gone, the backup succeeded.

I think the reason we didn't notice it before was indeed a not-fresh installation when the secrets were already copied to the namespace during other tests as we saw today in my local set up. Overall looks like the problem is something we missed during the multinamespaces transition.

recharte (approved):

PR #349, opened 2024-03-27
EVEREST-890 fix mysql pitr storage

oksana-grishchenko:
**PXC PITR storage fix**
---
**Problem:**
EVEREST-890

PXC can have a different storage for PITR. 
there was a regression introduced in https://github.com/percona/everest-operator/pull/317, where in the restores the PITR was configured to the same storage as the base backup.

**CHECKLIST**
---
**Jira**

**Tests**


controllers/databaseclusterrestore_controller.go
```diff
@@ -599,23 +599,11 @@ func (r *DatabaseClusterRestoreReconciler) genPXCPitrRestoreSpec(
...
-		}
-		//nolint:godox
-		// TODO: add support for Azure.
-	default:
-		return nil, fmt.Errorf("unsupported backup storage type %s for %s", backupStorage.Spec.Type, backupStorage.Name)
+		BackupSource: &pxcv1.PXCBackupStatus{
+			StorageName: pitrStorageName(*db.Spec.Backup.PITR.BackupStorageName),
+		},
+		Type: string(dataSource.PITR.Type),
+		Date: dataSource.PITR.Date.Format(everestv1alpha1.DateFormatSpace),
```
mayankshah1607, line 606: Does restoring to a new cluster still work?

mayankshah1607: The problem is that when restoring to a new cluster, it needs to use the same bucket as the source cluster..
With this change, it will use the bucket path that is configured for the new cluster, which won't work 🤔

oksana-grishchenko: I see, a good point, thanks. We need to configure the bucket explicitly, but for that to use the storage that is configured for pitr but not for the backup

oksana-grishchenko: I'll update it

oksana-grishchenko: I don't think so. We still need to know the source database.
The restore PITR section needs to use the same bucket that is defined in the source database cluster in the PITR configuration. 

Updated the codebase, tested for both restoring to the same and to the new cluster.

mayankshah1607 (approved):

PR #359, opened 2024-04-04
EVEREST-955 | Handle clean-up of backup data stored on s3

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-955

Allow deleting data from backup storage upon deletion of the backup.

**CHECKLIST**
---
**Jira**

**Tests**


controllers/databaseclusterbackup_controller.go
```diff
@@ -762,3 +794,27 @@ func backupStorageName(repoName string, cluster *everestv1alpha1.DatabaseCluster
 	}
 	return cluster.Spec.Backup.Schedules[scheduleInd-2].BackupStorageName, nil
 }
+
+// handleCleanup handles the cleanup of the DatabaseClusterBackup.
+func (r *DatabaseClusterBackupReconciler) handleStorageCleanup(
+	ctx context.Context,
+	dbcBackup *everestv1alpha1.DatabaseClusterBackup,
+	childBackup client.Object,
```
oksana-grishchenko, line 802: The name `childBackup` seems to be a bit confusing. For on-demand backups it's indeed like a child - first we create `dbb` CR then the upstream backup CR after it. But in case of scheduled backups the first one appeared is the upstream backup and then the `dbb` CR is created after it. 
So should it be something without the hierarchy in the name, e.g. `upstreamBackup`?

oksana-grishchenko (approved):
Looks good, shared couple of thoughts to consider

controllers/databaseclusterbackup_controller.go
mayankshah1607: Sure, `upstreamBackup` sounds better. Updated!

controllers/databaseclusterbackup_controller.go
```diff
@@ -684,20 +708,28 @@ func (r *DatabaseClusterBackupReconciler) reconcilePG(
...
-		return err
+		return false, err
+	}
+
+	if !backup.GetDeletionTimestamp().IsZero() {
+		// We can't handle this finalizer in PG yet, so we will simply remove it (if present).
+		// See: https://perconadev.atlassian.net/browse/K8SPG-538
+		if controllerutil.RemoveFinalizer(backup, common.DBBBackupStorageCleanupFinalizer) {
+			return true, r.Update(ctx, backup)
+		}
```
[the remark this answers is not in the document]
mayankshah1607: Sure, added a test

PR #360, opened 2024-04-05
EVEREST-960 | Add support for upgrading DB operators via DBEngine annotations

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-60

As a part of the component upgrades, we need to implement a mechanism for allowing upgrade of the DB operators.

**Solution:**
Setting an annotation on the DatabaseEngine CR with the desired version should trigger operator upgrades. The control loop simply checks for existing InstallPlans, approves them, and reconciles a status for the user.

**CHECKLIST**
---
**Jira**

**Tests**


controllers/databaseengine_controller.go
```diff
@@ -86,55 +104,164 @@ func (r *DatabaseEngineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		dbEngine.Status.State = everestv1alpha1.DBEngineStateInstalling
-	}
-	if ready { //nolint:nestif
-		dbEngine.Status.State = everestv1alpha1.DBEngineStateInstalled
-		matrix, err := r.versionService.GetVersions(engineType, dbEngine.Status.OperatorVersion)
-		if err != nil {
+
+		// Handle operator upgrade, if needed.
+		if done, err := r.handleOperatorUpgrade(ctx, dbEngine, ready); err != nil {
 			return ctrl.Result{}, err
```
michal-kralik, line 114: Would this block further dbEngine reconciliation in case of e.g. `errInstallPlanNotFound`?

controllers/databaseengine_controller.go
```diff
@@ -70,14 +82,20 @@ func (r *DatabaseEngineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	}); err != nil {
 		return ctrl.Result{}, err
 	}
 
+	defer func() {
+		if updErr := r.Status().Update(ctx, dbEngine); updErr != nil {
+			res = ctrl.Result{}
+			err = updErr
+		}
+	}()
```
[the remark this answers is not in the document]
mayankshah1607: Without this there were multiple calls to .Status().Update(), so just wanted to consolidate everything in a single place. But yeah, I get that this might make things confusing together with the named returns.

I've fixed it now

controllers/databaseengine_controller.go
mayankshah1607: Hmm, that's a good point.

I've fixed it now so that we fallthrough in case of this error.

e2e-tests/tests/features/dbengine_upgrade/02-install-olm-resources.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - script: |
+      cat <<EOF | kubectl apply -n $NAMESPACE -f -
+      apiVersion: operators.coreos.com/v1
+      kind: OperatorGroup
+      metadata:
+        name: everest-databases
+      spec:
+        targetNamespaces:
+        - $NAMESPACE
+      EOF
```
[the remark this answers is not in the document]
mayankshah1607: Because I needed to inject environment variables

e2e-tests/tests/features/dbengine_upgrade/31-assert.yaml
```diff
@@ -0,0 +1,14 @@
...
+apiVersion: everest.percona.com/v1alpha1
+kind: DatabaseEngine
+metadata:
+  name: percona-xtradb-cluster-operator
+spec:
+  type: pxc
+status:
+  status: installed
+  operatorVersion: 1.13.0
+
```
recharte, line 14: Should we also assert that the annotation has been removed?

controllers/databaseengine_controller.go
```diff
@@ -70,89 +83,201 @@ func (r *DatabaseEngineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	requeue := false
+	if done, err := r.handleOperatorUpgrade(ctx, dbEngine); err != nil {
+		if !errors.Is(err, errInstallPlanNotFound) {
+			return ctrl.Result{}, err
+		}
+		// We could not find the InstallPlan for the operator upgrade,
+		// so we will fallthrough since we'd still want the engine status to be updated.
+		// We will still requeue to check for the InstallPlan later.
+		requeue = true
+		logger.Error(err, "Failed to handle operator upgrade")
```
recharte, line 102: I think we could make this message be more specific and say we don't have the IP yet.

recharte (approved):
Great job Mayank :clap: just a couple of nitpick comments.

e2e-tests/tests/features/dbengine_upgrade/31-assert.yaml
mayankshah1607: Sure, updated :)

controllers/databaseengine_controller.go
mayankshah1607: Sounds good, updated..

PR #365, opened 2024-04-09
EVEREST-961 | List available operator upgrades in DBEngine status

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-961

As a part of the component upgrades, we want to present the user a list of versions of the operators they can upgrade to.

**Solution:**

We add a new `.status.pendingOperatorUpgrades` field in the DBEngine. The controller will list all InstallPlans that are pending, whose version is greater than the current installed version, and list it in this field.

**CHECKLIST**
---
**Jira**

**Tests**


.github/workflows/ci.yml
```diff
@@ -210,6 +210,8 @@ jobs:
           kubectl delete pvc --all-namespaces --all
           kubectl delete backupstorage --all-namespaces --all
           kubectl get ns -o name | grep kuttl  | awk -F '/' {'print $2'} | xargs --no-run-if-empty kubectl delete ns
+          kubectl delete ns operators olm || true
```
recharte, line 213: These namespaces are leftovers from previous releases, I don't think we need use them at all right now. Are these getting provisioned anywhere? Can't we do this cleanup manually just one time, instead?

.github/workflows/ci.yml
```diff
@@ -210,6 +210,8 @@ jobs:
...
+          kubectl get crd -o name | grep .coreos.com$ | awk -F '/' {'print $2'} | xargs --no-run-if-empty kubectl delete crd
```
recharte, line 214: Are we even installing OLM anywhere?

recharte (approved):
The core logic of the PR looks great, just left a couple of comments regarding the CI.

.github/workflows/ci.yml
mayankshah1607: There's a test step in the end to do a clean-up, but that happens only if the test succeeds. I added this in the clean-up step separately so that they're cleaned up even in case of failures

PR #368, opened 2024-04-11
EVEREST-919 Support path-style URL access for Backup Storages

recharte:
**Support path-style URL access for Backup Storages**
---
**Problem:**
EVEREST-919

S3 compatible APIs can support two flavors of bucket hosting styles:

[Path-style](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#path-style-access)

[Virtual-hosted–style](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#virtual-hosted-style-access)

Currently everest only supports virtual-hosted style but we should also support the path-style.

**Solution:**
PXC and PSMDB automatically select path-style if needed. PGBackrest needs a config option.

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/api/v1alpha1/backupstorage_types.go
+++ b/api/v1alpha1/backupstorage_types.go
@@ -45,6 +45,11 @@ type BackupStorageSpec struct {
 	//
 	// +kubebuilder:default:=true
 	VerifyTLS *bool `json:"verifyTLS,omitempty"`
+	// ForcePathStyle is set to use path-style URLs.
+	// If unspecified, the default value is false.
+	//
+	// +kubebuilder:default:=false
+	ForcePathStyle *bool `json:"forcePathStyle,omitempty"`
 
 	// Description stores description of a backup storage.
 	Description string `json:"description,omitempty"`
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -156,6 +156,11 @@ func (in *BackupStorageSpec) DeepCopyInto(out *BackupStorageSpec) {
 		*out = new(bool)
 		**out = **in
 	}
+	if in.ForcePathStyle != nil {
+		in, out := &in.ForcePathStyle, &out.ForcePathStyle
+		*out = new(bool)
+		**out = **in
+	}
 	if in.AllowedNamespaces != nil {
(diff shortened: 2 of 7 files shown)


mayankshah1607:
> PXC and PSMDB automatically select path-style if needed. PGBackrest needs a config option.

Are you sure that PSMDB automatically selects it? My understanding is that that PSMDBO needs to set the `ForcePathStyle` property [here](https://github.com/percona/percona-server-mongodb-operator/blob/main/pkg/psmdb/backup/pbm.go#L251-L259), but doesn't expose this in the CR (so this property is always set to false)

recharte:
> > PXC and PSMDB automatically select path-style if needed. PGBackrest needs a config option.
> 
> Are you sure that PSMDB automatically selects it? My understanding is that PSMDBO needs to set the `ForcePathStyle` property [here](https://github.com/percona/percona-server-mongodb-operator/blob/main/pkg/psmdb/backup/pbm.go#L251-L259), but doesn't expose this in the CR (so this property is always set to false)

:thinking: I don't see such an option on [that struct](https://pkg.go.dev/github.com/percona/percona-backup-mongodb/pbm/storage/s3#Conf).
Also, I have a minio instance set up that only supports path-style and I tested the mongodb backups and it worked fine :shrug:

mayankshah1607:
Okay, I think we were looking at different versions, I see it here on the main branch: https://github.com/percona/percona-backup-mongodb/blob/main/pbm/storage/s3/s3.go#L49

But if it works without it, thats great :)

mayankshah1607 (approved):
LGTM!

recharte:
> Okay, I think we were looking at different versions, I see it here on the main branch: https://github.com/percona/percona-backup-mongodb/blob/main/pbm/storage/s3/s3.go#L49
> 
> But if it works without it, thats great :)

I just dug a bit deeper and this is why it works, the default is for it to be enabled:
https://github.com/percona/percona-backup-mongodb/blob/883fd6318fd41bd48e4928160e4f83ce5e86f201/pbm/storage/s3/s3.go#L136-L138

The PSMDB operator doesn't expose this parameter, so we can't really do anything from the everest side but it works :sweat_smile:

PR #369, opened 2024-04-12
EVEREST-973 | add support for configuring CRVersion to use with operators

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-973

As a part of the operators upgrade, we would like to be able to edit/update the CRVersion. Currently the everest-operator does not allow changing the CRVersion once the cluster is installed.

**Solution:**
- Add an optional `.spec.crVersion` field in `DatabaseCluster`. This property represents the desired CRVersion. During installation, this field is unspecified, and will be set to the version of the operator.
- Add a `.status.crVersion` field in `DatabaseCluster`. This field represents the observed CRVersion.
- Add a `.status.reccomendedCrVersion` field in `DatabseCluster`. This property is set after the operator is upgraded (i.e, the operator and CR are at different versions). Before upgrading the operator to the next version, the CR must be updated to the recommended version by updating in the spec.
- This PR also includes a small refactor w.r.t how we were setting crVersion earlier.

**CHECKLIST**
---
**Jira**

**Tests**


oksana-grishchenko (approved):
LGTM. Thanks for the `clusterType` refactoring, the code looks cleaner👍

PR #374, opened 2024-04-23
EVEREST-1001 | Initial (local) backup for PG clusters should be handled as a normal DBBackup

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1001

When creating PG clusters, PG operator requires that we specify an initial backup location (repo1) for bootstrapping a cluster. Currently, everest-operator handles this by specifying a PVC source for this initial backup. This backup is hidden from the users and may not be used.

We want to let this backup be visible to the user.

**Solution:**
- Add a new BackupStorage type, `local'. This type may specify a PVC spec.
- Upon creation of a new PG Cluster, Everest operator creates a reserved BackupStorage of this local type. The PVC spec in this BackupStorage is used to configure the repo1 in pgBackRest.
- This BackupStorage is automatically cleaned up once the cluster is deleted
- Once the initial PGBackup (repo1) is created by PGO, we now allow creating the corresponding DBBackup object (earlier this was explicitly ignored)
- This DBBackup is protected from deletion using a finalizer.
- Restores are not allowed to use this backup.

**CHECKLIST**
---
**Jira**

**Tests**
~- [ ] Are unit tests added where appropriate?~


oksana-grishchenko (approved):
Looks good to me, great job👍 The db deletion shouldn't get stuck even in the foreground bc the backup finalizer is going to be deleted. The local backup storage should be deleted automatically.   

Should we protect the local storages and the initial backups from deletion attempts at the API level so the end user would get a clean error message?  

Sorry for taking so long to review, each time I opened the PR I thought "why would we need this if the backup can't be used for the restore?" but this time I think I figured it out from the parent ticket that it's just for the transparency sake. However it would be a bit weird that even though there is an initial backup we still ask the users to take another one to be able to restore😅

PR #380, opened 2024-04-26
EVEREST-732: use PG backup destination from PG

michal-kralik:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-732

Until https://jira.percona.com/browse/K8SPG-411 is done we work around not having the destination in the PerconaPGBackup CR by getting this info directly from S3.

This PR uses the destination from PG directly.


controllers/databaseclusterbackup_controller.go
```diff
@@ -758,21 +758,8 @@ func (r *DatabaseClusterBackupReconciler) reconcilePG(
...
-	if backup.Status.State == everestv1alpha1.BackupState(pgv2.BackupSucceeded) && backup.Status.Destination == nil {
-		db := &everestv1alpha1.DatabaseCluster{}
-		if err := r.Get(ctx, types.NamespacedName{
-			Name:      backup.Spec.DBClusterName,
-			Namespace: backup.Namespace,
-		}, db); err != nil {
-			logger.Error(err, "could not get database cluster ")
-		}
-		if err == nil {
-			backup.Status.Destination = r.getLastPGBackupDestination(ctx, backupStorage, db)
```
oksana-grishchenko, line 773: looks like this method is not used anymore. should we keep it?

michal-kralik: I removed it. Thanks

mayankshah1607 (approved):

PR #382, opened 2024-04-29
EVEREST-900 | Get rid of Secrets reconciler, move functionality to BackupStorage reconciler

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-900

We have a separate Secret reconciler for syncing the `UsedNamespaces` in BackupStorage objects. It's a bit of an anti-pattern because now we have 2 controllers managing the same Status. Instead this should be a part of the existing BackupStorage reconciler itself as a separate sub-routine/function.

The following changes are made in this PR:
- Get rid of the Secret reconciler
- Add a new step in the BackupStorage reconcile loop to List backupstorage Secrets and manage the UsedNamespaces status
- Add a watch for Secrets with the backupstorage label
- Remove unused code and minor improvements for maintainability

**CHECKLIST**
---
**Jira**

**Tests**


michal-kralik (approved):

controllers/backupstorage_controller.go
```diff
@@ -97,35 +103,78 @@ func (r *BackupStorageReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-				return ctrl.Result{}, err
-			}
+		if err := r.Update(ctx, defaultSecret); err != nil {
+			return ctrl.Result{}, err
 		}
 	}
 
-	logger.Info("updating secrets in used namespaces")
-	if err := r.handleSecretUpdate(ctx, bs, defaultSecret); err != nil {
+	// Add finalizer.
```
recharte, line 122: I don't think this comment adds anything :sweat_smile: the function is literally called `AddFinalizer` :see_no_evil: 
```suggestion
```

controllers/backupstorage_controller.go
```diff
@@ -97,35 +103,78 @@ func (r *BackupStorageReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		if err := r.Status().Update(ctx, bs); err != nil {
 			return ctrl.Result{}, err
 		}
 	}
 
+	logger.Info("updating secrets in used namespaces")
+	if err := r.handleSecretUpdate(ctx, bs, defaultSecret); err != nil {
+		return ctrl.Result{}, err
+	}
+
```
recharte, line 143: What's the reason for doing this after the status update?

recharte (commented):
Logic looks good, just trying to understand why `handleSecretUpdate` is now called after the status update.

controllers/backupstorage_controller.go
mayankshah1607: Sorry, I have the habit of leaving placeholder comments everywhere and then forgetting to remove it. :)

controllers/backupstorage_controller.go
mayankshah1607: `handleSecretUpdate` basically updates all the secrets in the `UsedNamespaces`. In order to do that, you need to refresh the list of UsedNamespaces first, which is why this step happens in the end.

For example, if a namespace has been deleted, and UsedNamespace is not updated first, then `handleSecretUpdate` will try to reconcile a non-existent Secret..

recharte: 🤔 I see a possible issue though. Let's say you add namespace A. You first update the status and then you try to create the secret. What if the secret creation fails, e.g. by lack of permissions? You are reporting in the status that namespace A has the secret but actually it does not...
I don't think it's correct to update the status before updating the secret, it can lead to race conditions...

mayankshah1607: But we don't create Secrets here. The secrets are created/copied in their respective DB controllers.

What's really happening here is:
- List all namespaces where the secret exists already (these is the `UsedNamespace`)
- Update the status
- Then inside `handleSecretUpdate`: for each used Namespace, update the secret.

Only if the Secret has already been created (which doesn't happen here), will it be a part of the `UsedNamespace`, and then be reconciled (only updated) later. If the Secret creation fails, then it won't even be a part of the status in the first place.

Does this make sense? Let me know if I'm still missing something..

mayankshah1607: TBH, I don't know why we need the `handleSecretUpdate` step here, since they are already being reconciled inside the DB controller. Maybe best to remove it from here to avoid confusion..

recharte: I think you're right, I'll double-check it nonetheless.
In the meantime maybe you know the answer to this one as well. What happens if someone manually changes the secret in a given namespace? Will the reconciliation be triggered on both controllers? Will the drift be automatically reconciled? By which controller?

recharte: Yeah I think this analysis is correct, so this PR can be merged as is, no issues here.

recharte (approved):

PR #390, opened 2024-05-02
EVEREST-1036 Delete DBRestore CR when DBC CR is deleted

recharte:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1036

The database is removed, but the restore still remains in the namespace.

**Cause:**
No mechanism to automatically delete the restore object was implemented

**Solution:**
Set DBR owner reference to DBC

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databaseclusterrestore_controller.go
+++ b/controllers/databaseclusterrestore_controller.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/AlekSi/pointer"
 	pgv2 "github.com/percona/percona-postgresql-operator/pkg/apis/pgv2.percona.com/v2"
 	psmdbv1 "github.com/percona/percona-server-mongodb-operator/pkg/apis/psmdb/v1"
 	pxcv1 "github.com/percona/percona-xtradb-cluster-operator/pkg/apis/pxc/v1"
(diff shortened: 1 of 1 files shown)


controllers/databaseclusterrestore_controller.go
```diff
@@ -166,6 +175,28 @@ func (r *DatabaseClusterRestoreReconciler) Reconcile(ctx context.Context, req ct
...
+	db *everestv1alpha1.DatabaseCluster,
+) error {
+	if len(restore.GetOwnerReferences()) == 0 {
+		restore.SetOwnerReferences([]metav1.OwnerReference{
+			{
+				APIVersion:         db.APIVersion,
+				Kind:               db.Kind,
+				Name:               db.Name,
+				UID:                db.UID,
+				BlockOwnerDeletion: pointer.ToBool(true),
```
oksana-grishchenko, line 190: A bit worried about this blocking. It would work fine when we'll have the `deleting` status and won't actually delete the cluster until all related backups and restores are deleted, but if we merge it now wouldn't it block the DB deletion for an non-obvious for the end user reason? I may miss something, but I mean - I try to delete the cluster, but it's not getting deleted bc there are still a restore exist, so I should delete the restore first. Wouldn't it be confusing?

recharte: The DBC deletion would only get blocked if it failed to delete the DBR but I don't see a reason why it would fail to delete the DBR, there is nothing that would block the deletion of the DBR, e.g. there are no finalizers.

oksana-grishchenko: Ok, I see, looks like I've misinterpret the field. Resolved.

oksana-grishchenko (approved):

PR #392, opened 2024-05-06
EVEREST-874 Deleting state

oksana-grishchenko:
**Add deleting state**
---
**Problem:**
EVEREST-874

Currently, as soon as a user deletes a DB, the DBC CR gets removed. This PR introduces the deleting state that will remain until all resources belonging to that DB get removed.

**CHECKLIST**
---
**Jira**

**Tests**


oksana-grishchenko:
The CI tests failure seems to be not related. Figuring out if it can be improved, but it's not a blocker for the PR review.
Screenshot 2024-05-07 at 09 45 46

michal-kralik (approved):

controllers/providers/pg/provider.go
```diff
@@ -111,6 +111,9 @@ func (p *Provider) Status(ctx context.Context) (everestv1alpha1.DatabaseClusterS
 
 	status := p.DB.Status
 	status.Status = everestv1alpha1.AppState(pg.Status.State)
+	if !p.DB.DeletionTimestamp.IsZero() {
+		status.Status = everestv1alpha1.AppStateDeleting
+	}
```
mayankshah1607, line 116: Does this code ever get executed?

Based on my understand, in the `reconcileDB()` function in the outer controller, we do an early return if the deletionTimestamp is not zero. Which means `Status()` doesn't get called

controllers/common/consts.go
```diff
@@ -64,6 +64,9 @@ const (
 
 	// DBBackupCleanupFinalizer is the finalizer for cleaning up DatabaseClusterBackup.
 	DBBackupCleanupFinalizer = "everest.percona.com/dbb-cleanup"
+
+	// DBDeleteFinalizer is the finalizer for deleting the DatabaseCluster related resources.
+	DBDeleteFinalizer = "everest.percona.com/db-delete"
```
mayankshah1607, line 69: Do we really need this finalizer? It seems to me that the only purpose it serves it to ensure that we see a 'deleting' state at least once.

 Currently the DB deletion may be blocked only due to the `dbb-delete` finalizer.

I think we can show a deleting state even without the use of this finalizer:

```diff
diff --git a/controllers/databasecluster_controller.go b/controllers/databasecluster_controller.go
index fb0313b..ff3927e 100644
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -133,7 +133,8 @@ func (r *DatabaseClusterReconciler) reconcileDB(
                if done, err := p.Cleanup(ctx, db); err != nil {
                        return ctrl.Result{}, err
                } else if !done {
-                       return ctrl.Result{Requeue: true}, nil
+                       db.Status.Status = everestv1alpha1.AppStateDeleting
+                       return ctrl.Result{Requeue: true}, r.Status().Update(ctx, db)
                }
                return ctrl.Result{}, nil
        }
```

With this change, the outer controller simply checks if the inner Cleanup is complete. If not, we update the Status to deleting. We already have finalizers in place that will block the deletion until the clean-up of the respective child resources.

What do you think?

mayankshah1607: This way, the inner Cleanup is really only responsible for the actual clean-up, and the outer controller updates the state to Deleting if the clean-up is not complete.

Since we already have finalizers that will prevent deletion, we don't need this new finalizer

controllers/common/helper.go
```diff
@@ -698,3 +698,51 @@ func GetRecommendedCRVersion(
...
+) (bool, error) {
+	backupList, err := ListDatabaseClusterBackups(ctx, c, dbName, dbNs)
+	if err != nil {
+		return false, err
+	}
+
+	restoresList, err := ListDatabaseClusterRestores(ctx, c, dbName, dbNs)
+	if err != nil {
+		return false, err
+	}
```
mayankshah1607, line 742: I don't know if we should be checking the restores at all, since the deletion of restores is controlled using OwnerReferences.

Since the OwnerReference contains the `blockOwnerDeletion` flag, instead of explicitly checking for restores, we should perform a foreground deletion from the Everest API. That way, Kubernetes will guarentee the the DatabaseCluster stays as long as the restores are not deleted.

mayankshah1607: Here's an example of a foreground deletion we do for backups: https://github.com/percona/everest/blob/7ecd61206826ff849839bd1bb25ce87107870556/api/database_cluster_backup.go#L105

controllers/common/consts.go
mayankshah1607: Or even better if we're doing the foreground deletion I mentioned in https://github.com/percona/everest-operator/pull/392#discussion_r1592129708 :

```diff
diff --git a/controllers/databasecluster_controller.go b/controllers/databasecluster_controller.go
index fb0313b..9f56419 100644
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -130,12 +130,14 @@ func (r *DatabaseClusterReconciler) reconcileDB(
 ) (ctrl.Result, error) {
        // Handle any necessary cleanup.
        if !db.GetDeletionTimestamp().IsZero() {
-               if done, err := p.Cleanup(ctx, db); err != nil {
+               done, err := p.Cleanup(ctx, db)
+               if err != nil {
                        return ctrl.Result{}, err
-               } else if !done {
-                       return ctrl.Result{Requeue: true}, nil
                }
-               return ctrl.Result{}, nil
+               db.Status.Status = everestv1alpha1.AppStateDeleting
+               return ctrl.Result{
+                       Requeue: !done,
+               }, r.Status().Update(ctx, db)
        }

        // Set metadata.
```

controllers/providers/pg/provider.go
oksana-grishchenko: In this PR the Status() is called separately in the cleanup (like this https://github.com/percona/everest-operator/blob/977c8db0e5b95d8b93021f179be3a5a42c31a7a0/controllers/providers/pg/provider.go#L143-L146) and then the status got applied inside `HandleDBDeletion`.

But let me rethink the whole approach according to your feedback regarding finalizers. Thank you!

controllers/common/consts.go
oksana-grishchenko: Thank you for the feedback, makes sense, let me rework the approach👍

controllers/common/helper.go
mayankshah1607: This will make sure that Kubernetes keeps the DBC in until every child resource is deleted, not just the restores

oksana-grishchenko: There is no need for listing the restores indeed if the foreground deletion is used. Thank you for the idea, I love it how concise the code has become👍

controllers/databasecluster_controller.go
```diff
@@ -133,7 +134,8 @@ func (r *DatabaseClusterReconciler) reconcileDB(
 		if done, err := p.Cleanup(ctx, db); err != nil {
 			return ctrl.Result{}, err
 		} else if !done {
-			return ctrl.Result{Requeue: true}, nil
+			db.Status.Status = everestv1alpha1.AppStateDeleting
+			return ctrl.Result{Requeue: true}, r.Status().Update(ctx, db)
```
mayankshah1607, line 138: I think we should set the status regardless of the value of `done`. What happens if `Cleanup` returns `done=true` the first time it is called, but the foregroundDeletion is still pending. In that case, we may never see a `deleting` state on the DBC, even though it is actually being deleted.

```diff
diff --git a/controllers/databasecluster_controller.go b/controllers/databasecluster_controller.go
index fb0313b..9f56419 100644
--- a/controllers/databasecluster_controller.go
+++ b/controllers/databasecluster_controller.go
@@ -130,12 +130,14 @@ func (r *DatabaseClusterReconciler) reconcileDB(
 ) (ctrl.Result, error) {
        // Handle any necessary cleanup.
        if !db.GetDeletionTimestamp().IsZero() {
-               if done, err := p.Cleanup(ctx, db); err != nil {
+               done, err := p.Cleanup(ctx, db)
+               if err != nil {
                        return ctrl.Result{}, err
-               } else if !done {
-                       return ctrl.Result{Requeue: true}, nil
                }
-               return ctrl.Result{}, nil
+               db.Status.Status = everestv1alpha1.AppStateDeleting
+               return ctrl.Result{
+                       Requeue: !done,
+               }, r.Status().Update(ctx, db)
        }

        // Set metadata.
```

mayankshah1607 (approved):
Thanks for making all the changes. Code is very concise now!

Just one more small suggestion, otherwise LGTM. 🙂

api/v1alpha1/databasecluster_types.go
```diff
@@ -47,6 +47,8 @@ const (
 	AppStateError AppState = "error"
 	// AppStateRestoring is a restoring state.
 	AppStateRestoring AppState = "restoring"
+	// AppStateDeleting is a restoring state.
+	AppStateDeleting AppState = "deleting"
```
mayankshah1607, line 51:
```suggestion
	// AppStateDeleting is a deleting state.
	AppStateDeleting AppState = "deleting"
```

controllers/databasecluster_controller.go
oksana-grishchenko: Right, thanks, updated👍

PR #396, opened 2024-05-08
EVEREST-1054 | Fix PG restores for S3 compatible storage locations 

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1054

PG restores do not work when using a backup storage location such as Minio.

**Cause:**
The restore issues are only caused when setting `forcePathStyle` and/or `verifyTLS=false`

#### 1. When restoring to same cluster:
The backup controller lists objects in the S3 storage to build the backup destination path. This S3 API call was failing as the client was not configured for skipping TLS checks or using path-style URIs.

#### 2. When restoring to new cluster:
When performing restores to a new cluster, we did not configure pgBackRest to skip TLS checks or force path-style URIs, depending on the backup storage.

**Solution:**
- Enforce TLS settings and URI style in S3 client while listing S3 objects in the backup controller
- Similarly, configure pgBackRest settings correctly when restoring to a new cluster.

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/providers/pg/applier.go
+++ b/controllers/providers/pg/applier.go
@@ -589,7 +589,8 @@ func (p *applier) genPGDataSourceSpec() (*crunchyv1beta1.DataSource, error) {
 func getPGRestoreOptions(
 	dataSource everestv1alpha1.DataSource,
 	backupStorage *everestv1alpha1.BackupStorage,
-	backupBaseName, repoName string) ([]string, error) {
+	backupBaseName, repoName string,
+) ([]string, error) {
 	options := []string{
 		"--set=" + backupBaseName,
 	}
--- a/controllers/providers/pg/applier.go
+++ b/controllers/providers/pg/applier.go
@@ -549,7 +549,7 @@ func (p *applier) genPGDataSourceSpec() (*crunchyv1beta1.DataSource, error) {
 	}
 
 	repoName := "repo1"
-	options, err := getPGRestoreOptions(*database.Spec.DataSource, backupBaseName)
+	options, err := getPGRestoreOptions(*database.Spec.DataSource, backupStorage, backupBaseName, repoName)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


controllers/providers/pg/applier.go
```diff
@@ -588,11 +588,21 @@ func (p *applier) genPGDataSourceSpec() (*crunchyv1beta1.DataSource, error) {
...
+	backupBaseName, repoName string) ([]string, error) {
 	options := []string{
 		"--set=" + backupBaseName,
 	}
 
+	if pointer.Get(backupStorage.Spec.ForcePathStyle) {
+		options = append(options, "--"+fmt.Sprintf(pgBackRestStorageForcePathTmpl, repoName)+"="+pgBackRestStoragePathStyle)
+	}
+	if !pointer.Get(backupStorage.Spec.VerifyTLS) {
+		options = append(options, "--no-"+fmt.Sprintf(pgBackRestStorageVerifyTmpl, repoName))
```
recharte, line 601: Are you sure this works? I couldn't find this in the PGBackrest docs...
I could only find `repo1-storage-verify-tls=n`
https://pgbackrest.org/configuration.html#section-repository/option-repo-storage-verify-tls

controllers/providers/pg/applier.go
```diff
@@ -586,11 +586,21 @@ func (p *applier) genPGDataSourceSpec() (*crunchyv1beta1.DataSource, error) {
...
+	}
```
recharte, line 602: I'm a bit confused why we need this at all. Should [this](https://github.com/percona/everest-operator/blob/939c65dc1771cef0b9ef2a645f173fa15f219809/controllers/providers/pg/applier.go#L1116-L1123) already take care of it? :thinking:

controllers/providers/pg/applier.go
mayankshah1607: I wasn't able to get the anchor tag for this, but you'll find it under this section: https://pgbackrest.org/command.html#command-restore

Screenshot 2024-05-08 at 6 36 33 PM

controllers/providers/pg/applier.go
mayankshah1607: That's what I thought at first, but it doesn't. The snippet you provided only takes care of the backups.
When you restore to a new cluster, those settings are lost. So you need to set them again

mayankshah1607: This `getPGRestoreOptions` function is called only when we specify the `DataSource` field. Whereas in the snipped provided above, those are used only for backups and restoring to the same cluster. So I don't think there should be any conflicts/issues..

mayankshah1607: BTW, I did a run through of the backup/restore scenario with these changes, and it worked as expected

recharte: The thing that's concerning me is the hardcoded `repo1` (I know it was already like this).

recharte: Okay so I just dug a bit deeper on this and I think everything is fine. Like you said this is completely isolated from the actual DB. The repos that we define here with the `handlePGDataSourceS3` have absolutely nothing to do with the ones defined outside the `DataSource` field.

recharte (approved):

oksana-grishchenko (approved):

PR #398, opened 2024-05-09
EVEREST-979 | Reverse the behaviour of storage finalizers

mayankshah1607:
Earlier, we always removed the S3 finalizer from Backups, and conditionally added it from the Everest API, if specified by the user.

This commit reverses this behaviour, i.e, we ensure the S3 finalizer is always present. With this change, S3 is always cleaned up, unless the user chooses to skip it from the Everest API.

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-979

S3 is not being cleaned up after backups are deleted due to retention policy settings.

**Cause:**

By default, we always remove the S3 finalizer from the upstream backup objects. Due to this, when the operator deletes the backup, they're not cleaned-up from S3.

**Solution:**

Reverse the behaviour of the S3 finalizer. I.e, we now ensure that the finalizer is always present. We only remove it if the user chooses to explicitly keep the data during backup/cluster deletion. 

**CHECKLIST**
---
**Jira**

**Tests**


michal-kralik (approved):

oksana-grishchenko (commented):
LGTM, two things to mention:
1. I assume there will be some more changes bc of the failing tests (seems like the failure is related to the changes), that's why leaving no approval so far although the logic looks fine. 
2. Maybe the db foreground deletion is the reason of the failures in some tricky way, I faced some similar issues earlier with the db clusters deletion. For the db cluster there is [this code](https://github.com/percona/everest-operator/blob/ba7373f4e212fddc2e732120c7e1fbe73bf946b8/controllers/databasecluster_controller.go#L785-L787) added to the `ensureFinalizers` method to prevent the db finalizers re-appering after the db is deleted (when the delete timestamp is not empty).

mayankshah1607:
Thanks @oksana-grishchenko , the failures were related to the tests added in this PR - https://github.com/percona/everest-operator/pull/359

They've been fixed now

oksana-grishchenko (approved):

PR #399, opened 2024-05-13
EVEREST-1061 | backups should not be reconciled after they've succeeded

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1061

Backups should not be reconciled after they've completed.

**Why is this needed?**

When retentionCopies policy is set, the underlying operators will delete their backup CR. Everest operator sets a bi-directional owner reference between the dbbackup and the upstream backup CR. This could lead to a race condition where the upstream operator deletes the backup CR (due to retention settings) but the Everest operator re-creates it.

This can be fixed if the everest operator does not reconcile backups that have completed.

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/api/v1alpha1/databaseclusterbackup_types.go
+++ b/api/v1alpha1/databaseclusterbackup_types.go
@@ -16,6 +16,9 @@
 package v1alpha1
 
 import (
+	pgv2 "github.com/percona/percona-postgresql-operator/pkg/apis/pgv2.percona.com/v2"
+	psmdbv1 "github.com/percona/percona-server-mongodb-operator/pkg/apis/psmdb/v1"
+	pxcv1 "github.com/percona/percona-xtradb-cluster-operator/pkg/apis/pxc/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -69,6 +72,13 @@ type DatabaseClusterBackupList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []DatabaseClusterBackup `json:"items"`
+}
+
+// IsComplete returns true if the backup has completed.
+func (b DatabaseClusterBackup) IsComplete() bool {
+	return b.Status.State == BackupState(pxcv1.BackupSucceeded) ||
+		b.Status.State == BackupState(pgv2.BackupSucceeded) ||
+		b.Status.State == BackupState(psmdbv1.BackupStateReady)
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


api/v1alpha1/databaseclusterbackup_types.go
```diff
@@ -71,6 +74,13 @@ type DatabaseClusterBackupList struct {
 	Items           []DatabaseClusterBackup `json:"items"`
 }
 
+// IsComplete returns true if the backup has completed.
+func (b DatabaseClusterBackup) IsComplete() bool {
+	return b.Status.State == BackupState(pxcv1.BackupSucceeded) ||
+		b.Status.State == BackupState(pgv2.BackupSucceeded) ||
+		b.Status.State == BackupState(psmdbv1.BackupStateReady)
```
recharte, line 81: What about the failed states? I think they are also terminal states.

mayankshah1607: right, thanks. Updated

controllers/databaseclusterbackup_controller.go
```diff
@@ -115,6 +115,11 @@ func (r *DatabaseClusterBackupReconciler) Reconcile(ctx context.Context, req ctr
 		}
 		return reconcile.Result{}, err
 	}
+
+	if backup.IsComplete() && backup.GetDeletionTimestamp().IsZero() {
+		return reconcile.Result{}, err
```
recharte, line 121: We don't have an `err` to return here

recharte: Actually it makes no difference but I'd rather us be explicit.
```suggestion
		return reconcile.Result{}, nil
```

recharte (approved):

controllers/databaseclusterbackup_controller.go
```diff
@@ -115,6 +115,12 @@ func (r *DatabaseClusterBackupReconciler) Reconcile(ctx context.Context, req ctr
 		}
 		return reconcile.Result{}, err
 	}
+
+	if (backup.HasSucceeded() || backup.HasFailed()) &&
```
recharte, line 119: I just remembered something...
If I'm not mistaken I think I've seen in the past the status move temporarily to an error state and then getting back to in progress and eventually succeed. I just not sure in which resource this was, DBC, DBB or DBR... :thinking:

recharte: I'm now slightly worried about having added the `backup.HasFailed()`. I need to test this thoroughly...

mayankshah1607: I can't be 100% sure, but I think it was the DBRestore object..

recharte: Yeah I think you're right. Let's not block the PR for this, I'll do some more tests to confirm...

PR #400, opened 2024-05-13
EVEREST-1064 | psmdb-backups should not be adopted by dbbackups until a status is observed

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1064

Duplicate backups are created when specifying a schedule for PSMDB.

**Cause:**
When a backup schedule is specified, PSMDBO automatically creates a psmdb-backup. Upon creation, the backup process starts. However, Everest then modifies the psmdb-backup object by setting an OwnerReference, which triggers another backup run for the same psmdb-backup object, resulting in duplicate backups.

**Solution:**
This is a known issue with PSMDBO, and a fix is expected. As a temporary workaround, Everest should wait for the psmdb-backup to progress beyond "waiting" to be able to adopt the object.

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databaseclusterbackup_controller.go
+++ b/controllers/databaseclusterbackup_controller.go
@@ -409,6 +409,13 @@ func (r *DatabaseClusterBackupReconciler) tryCreatePSMDB(ctx context.Context, ob
 		return err
 	}
 
+	// Wait until the backup has progressed beyond the waiting state,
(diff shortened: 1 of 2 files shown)


recharte:
I like this solution :+1:

recharte (approved):

PR #415, opened 2024-06-07
EVEREST-1137 | [PXC] Fix how external access is enabled based on changes from PXCO 1.14.0

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1137

After upgrading to PXC Operator 1.14.0, we no longer see the public DNS/IP Address on the Host when enabling external access.

**Related pull requests**

- https://github.com/percona/percona-xtradb-cluster-operator/pull/1538

**Cause:**
In PXCO 1.14.0, new fields were added which need to be configured in order to enable external access for DB services. This PR accounts for those changes so that DBs are exposed correctly when using PXCO 1.14.0

**Solution:**
Set the newly added `Expose` fields. We also preserve the older (deprecated) configuration in order to maintain backward compatibility with 1.13.0.

**CHECKLIST**
---
**Jira**


recharte:
@mayankshah1607 Why did the old way stopped working? We asked about this to the cloud team and they said that they would ensure backward compatibility and we wouldn't need to do these changes in this release.

mayankshah1607:
@recharte The old way works as long as the CRVersion is 1.13.0 - we still have backward compatibility in that sense. But once you switch to 1.14.0, it expects that the newly added expose fields are set. The old fields are not honoured when CRVersion is 1.14.0. 

These fields which we use have been deprecated in 1.14.0 and we will have to clean up once 1.13.0 is not supported - https://github.com/percona/percona-xtradb-cluster-operator/pull/1538/files#diff-c8e4948efd302342743e1727d83baea8b575b085896d9f77199839ffd4a951abR434-L433

recharte:
> @recharte The old way works as long as the CRVersion is 1.13.0 - we still have backward compatibility in that sense. But once you switch to 1.14.0, it expects that the newly added expose fields are set. The old fields are not honoured when CRVersion is 1.14.0.
> 
> These fields which we use have been deprecated in 1.14.0 and we will have to clean up once 1.13.0 is not supported - https://github.com/percona/percona-xtradb-cluster-operator/pull/1538/files#diff-c8e4948efd302342743e1727d83baea8b575b085896d9f77199839ffd4a951abR434-L433

Hmmm :thinking: I see... For some reason I assumed that even within a given CR version they would support both but that doesn't make any sense. What makes sense (which is what they do) is that operator version 1.14 supports CR versions 1.13 and 1.14 but if you use CR version 1.14 you must have the correct schema without the deprecated fields.
Thanks for clearing it up.

recharte (approved):
Looks good but please add a comment in the code that sets the deprecated fields to remind us to remove them once we add support for 1.15

go.mod
```diff
@@ -133,3 +133,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/go-jose/go-jose/v4 => github.com/go-jose/go-jose/v4 v4.0.1
```
recharte, line 137: Why?

recharte: Are the last two commits (updating cert-manager and this change) an attempt to fix the vulnerability detected by snyk?

mayankshah1607: Yes

Even this was an attempt to fix it. I will be removing it before merging

PR #418, opened 2024-06-10
EVEREST-1139 Fix stuck PG deletion

oksana-grishchenko:
**Fix stuck PG deletion**
---
This PR is converted back to draft in favour of the core problem solution PR 
- https://github.com/percona/everest-operator/pull/419
---

**Problem:**
EVEREST-1139
Deletion of PG database cluster may get stuck if there are several scheduled backups taken. 

**Cause:**
In scheduled backups there is this circular `OwnerReferences` dependency between `dbb` and `pg-backup` and since the foreground deletion (which gets propagated to the owned resources) is used, the dbb and pg-backup at some point got to the situation they own each other but they need to wait until the owned resource is deleted. 
Screenshot 2024-06-10 at 10 57 26

**Solution:**
Ideal solution would be to rethink the circular OwnerReference approach, however it requires a significant redesign and retesting, so this PR only fixes the symptoms. 
Since the problem doesn't show up for `pxc` and `psmdb` because an additional reconciliation requeue happens when the storage cleanup finalizer is deleted from the upstream backup, this PR adds an additional requeue for the PG too. 

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databaseclusterbackup_controller.go
+++ b/controllers/databaseclusterbackup_controller.go
@@ -736,6 +736,19 @@ func (r *DatabaseClusterBackupReconciler) reconcilePG(
 		if controllerutil.RemoveFinalizer(backup, everestv1alpha1.DBBackupStorageProtectionFinalizer) {
 			return true, r.Update(ctx, backup)
 		}
+
+		// For the scheduled backups, need to requeue reconciliation one more time
+		// because the circular OwnerReferences dependency between dbb and pg-backup
+		// along with the foreground deletion (which gets propagated to the owned resources)
+		// may make the deletion stuck. The backward OwnerReference "pg-backup owns dbb" can be
+		// safely deleted since the dbb already has the deletion timestamp.
+		// The problem doesn't show up for pxc and psmdb bc the additional reconciliation requeue happens
+		// when the storage cleanup finalizer is deleted from the upstream backup.
+		if len(backup.GetOwnerReferences()) > 0 {
+			// Custom resources may have only one OwnerReference, so cleaning up the list.
+			backup.SetOwnerReferences([]metav1.OwnerReference{})
+			return true, r.Update(ctx, backup)
+		}
 	}
 
 	backupStorage := &everestv1alpha1.BackupStorage{}


controllers/databaseclusterbackup_controller.go
```diff
@@ -736,6 +736,19 @@ func (r *DatabaseClusterBackupReconciler) reconcilePG(
...
 			return true, r.Update(ctx, backup)
 		}
+
+		// For the scheduled backups, need to requeue reconciliation one more time
+		// because the circular OwnerReferences dependency between dbb and pg-backup
+		// along with the foreground deletion (which gets propagated to the owned resources)
+		// may make the deletion stuck. The backward OwnerReference "pg-backup owns dbb" can be
+		// safely deleted since the dbb already has the deletion timestamp.
+		// The problem doesn't show up for pxc and psmdb bc the additional reconciliation requeue happens
+		// when the storage cleanup finalizer is deleted from the upstream backup.
```
mayankshah1607, line 746: I'm still a bit confused why this worked for PSMDB and PXC.. even in the additional requeue we did there, we never explicitly removed the OwnerRefs. So why doesn't that lead to a deadlock? Does the `foregroundDeletion` finalizer not get propagated there? 🤔

oksana-grishchenko: You're right, we never delete the `OwnerReference` from `pxc` and `psmdb`, the thing is that we only need _one more requeue_, and for that we need to change something in the CR so we could rely on the change as an indicator. It is safe to delete the `OwnerReference` from the `dbb` because the OwnerReference used there only to delete dbb in case the retention copies number is exceeded, but since the dbb already has the deletion timestamp in place, it's safe to delete the reference. 

We can also rely on some other change to be able to make the requeue only once. Removing the OwnerReference is just one option which seems to be the safest one. 

another option would be instead of the new block that this PR introduces just to add
```
return true, nil
``` 
so it would work like this: if the `dbb` has the deletion timestamp in place, the reconciliation will be requeued endless until the CR is deleted. It adds an unnecessary overhead and in general such endless loops can go out of control easily.

oksana-grishchenko: On a second thought, let me try to remove the circular dependency in a separate PR. I'm going to remove the direction "upstream backup owns everest backup" and let a watcher track the upstream deletion so that the related dbb would be deleted too.

michal-kralik (approved):

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: Created the PR to remove circular OwnerReferences https://github.com/percona/everest-operator/pull/419 

Moving the current PR back to draft and will close it once we accept the OwnerReferences removing

oksana-grishchenko:
Closing the PR: fixed in https://github.com/percona/everest-operator/pull/419

PR #419, opened 2024-06-10
EVEREST-1139 Remove circular OwnerReferece from scheduled backups

oksana-grishchenko:
**Remove circular OwnerReferece from scheduled backups**
---
**Problem:**
EVEREST-1139

The problem and the reasons are described in this PR which supposed to fix the symptoms and not needed anymore:
- https://github.com/percona/everest-operator/pull/418 

This PR fixes the core issue - the circular dependency in the scheduled backups.

Tested the `dbb`, `pxc-backup`, `psmdb-backup`, `pg-backup` resources creation and deletion for the scheduled and on-demand backups (including the retention copies) - works as expected. 

**CHECKLIST**
---
**Jira**

**Tests**


controllers/databaseclusterbackup_controller.go
```diff
@@ -252,27 +251,70 @@ func (r *DatabaseClusterBackupReconciler) SetupWithManager(mgr ctrl.Manager, sys
...
 
+func (r *DatabaseClusterBackupReconciler) watchHandler(creationFunc func(ctx context.Context, obj client.Object) error) handler.Funcs {
+	return handler.Funcs{
+		CreateFunc: func(ctx context.Context, e event.CreateEvent, q workqueue.RateLimitingInterface) {
+			r.tryCreateDBBackups(ctx, e.Object, creationFunc)
+			q.Add(reconcileRequestFromObject(e.Object))
+		},
+		UpdateFunc: func(ctx context.Context, e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+			// remove the r.tryCreateDBBackups call below once https://perconadev.atlassian.net/browse/K8SPSMDB-1088 is fixed.
+			r.tryCreateDBBackups(ctx, e.ObjectNew, creationFunc)
```
oksana-grishchenko, line 262: This attempt to create a `dbb` resource appears in the `UpdateFunc` bc of the [workaround](https://github.com/percona/everest-operator/blob/4f738f6/controllers/databaseclusterbackup_controller.go#L407-L412). 
The workaround basically exits the `dbb` resource creation for the `psmdb` scheduled backups if the status of the `psmdb-backup` is empty, and in that case there is no other chance for dbb to be created. So the creation attempt is added to the `UpdateFunc` too. 

@mayankshah1607 since you've worked on the workaround please let me know if you have other ideas of how to handle the limitation.

mayankshah1607 (approved):
LGTM, thanks!

controllers/databaseclusterbackup_controller.go
mayankshah1607: Yeah, this makes sense. I don't think I have a better idea than this..

PR #428, opened 2024-06-14
EVEREST-1141 Backup storage files deletion fix

oksana-grishchenko:
**Backup storage files cleanup fix**
---
**Problem:**
EVEREST-1141

**Problems**
1. Backup files were not deleted from s3 after the `dbc` was deleted
2. PSMDB limitation: can't delete the last backup while PITR is on

**Cause:**
Since the upstream cluster had its owner reference pointing to the `dbc`, once the dbc was deleted the upstream cluster was also deleted, but it caused the upstream backups cleanup finalizers (like `delete-backup` for psmdb) failure because the upstream cluster wasn't there anymore. 

**Solution/what's changed:**
1. The upstream clusters are cleaned up now explicitly _after the backups deletion_ by using the new  "everest.percona.com/upstream-cluster-cleanup" finalizer. The owner reference "dbc owns upstream cluster" is deleted.
3. Since there is no circular dependencies between `dbb` and upstream backups anymore, the `dbb` is now owned by `dbc`, so that when the `dbb` is deleted, the controller doesn't delete the `dbb`s explicitly but just waits for them to be deleted according to the owner references.  
4. Handled the psmdb limitation (unable to delete the last backup while PITR is on) by disabling pitr in `psmdb` before `dbb` deletion.

**CHECKLIST**
---
**Jira**

**Tests**


controllers/databaseclusterbackup_controller.go
```diff
@@ -372,10 +373,21 @@ func (r *DatabaseClusterBackupReconciler) tryCreatePG(ctx context.Context, obj c
...
+func dbOwnerRefMeta(db *everestv1alpha1.DatabaseCluster) []metav1.OwnerReference {
+	return []metav1.OwnerReference{{
+		APIVersion:         everestAPIVersion,
+		Kind:               databaseClusterKind,
+		Name:               db.Name,
+		UID:                db.UID,
+		BlockOwnerDeletion: pointer.ToBool(true),
+	}}
+}
+
```
mayankshah1607, line 390: We can use `controllerutil.SetControllerReference` directly, then we can remove this helper function. See example: https://github.com/percona/everest-operator/blob/main/controllers/monitoringconfig_controller.go#L96

controllers/providers/pg/provider.go
```diff
@@ -134,7 +134,14 @@ func (p *Provider) Status(ctx context.Context) (everestv1alpha1.DatabaseClusterS
 
 // Cleanup runs the cleanup routines and returns true if the cleanup is done.
 func (p *Provider) Cleanup(ctx context.Context, database *everestv1alpha1.DatabaseCluster) (bool, error) {
-	return common.HandleDBBackupsCleanup(ctx, p.C, database)
+	done, err := common.HandleDBBackupsCleanup(ctx, p.C, database)
+	if err != nil {
+		return false, err
+	}
+	if !done {
+		return true, nil
```
mayankshah1607, line 142: Shouldn't it return `false, nil`? (The comment says `.. returns true if cleanup is done`)

controllers/databaseclusterbackup_controller.go
```diff
@@ -414,6 +426,13 @@ func (r *DatabaseClusterBackupReconciler) tryCreatePXC(ctx context.Context, obj
 		databaseClusterNameLabel: pxcBackup.Spec.PXCCluster,
 		fmt.Sprintf(backupStorageNameLabelTmpl, pxcBackup.Spec.StorageName): backupStorageLabelValue,
 	}
+	cluster := &everestv1alpha1.DatabaseCluster{}
+	err = r.Get(ctx, types.NamespacedName{Name: pxcBackup.Spec.PXCCluster, Namespace: pxcBackup.Namespace}, cluster)
+	if err != nil {
+		return err
+	}
+	backup.SetOwnerReferences(dbOwnerRefMeta(cluster))
```
mayankshah1607, line 434: Shouldn't we also set an ownerRef for on-demand backups? If I understand correctly, this will ensure that ownerRef is set only for scheduled backups

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: Thanks, updated

controllers/providers/pg/provider.go
oksana-grishchenko: Right, it should've, thank you. However the code is not relevant anymore since the `dbb-cleanup` finalizer is deleted.

controllers/common/consts.go
```diff
@@ -62,6 +62,9 @@ const (
 	// DBBackupCleanupFinalizer is the finalizer for cleaning up DatabaseClusterBackup.
 	DBBackupCleanupFinalizer = "everest.percona.com/dbb-cleanup"
```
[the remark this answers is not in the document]
oksana-grishchenko: Thanks for pointing that out, it's updated. I was keeping the finalizer to check that the backups are deleted, but there is no need for a separate finalizer for it anymore indeed, so the backups deletion is checked in the `HandleUpstreamClusterCleanup`

controllers/databaseclusterbackup_controller.go
oksana-grishchenko: Great catch, thank you, updated the metadata reconciliation so that the on-demand backups would have the ownerref too.

controllers/databasecluster_controller.go
```diff
@@ -659,6 +656,49 @@ func (r *DatabaseClusterReconciler) initWatchers(controller *builder.Builder) {
...
+				{
+					NamespacedName: types.NamespacedName{
+						Name:      obj.GetName(),
+						Namespace: obj.GetNamespace(),
+					},
+				},
+			}
+		}),
+		builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
+	)
```
mayankshah1607, line 701: Can be simplified using `EnqueueRequestForObject` (it achieves the same thing)
```suggestion
	controller.Watches(
		&psmdbv1.PerconaServerMongoDB{},
		&handler.EnqueueRequestForObject{},
		builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
	)
	controller.Watches(
		&pgv2.PerconaPGCluster{},
		&handler.EnqueueRequestForObject{},
		builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
	)
	controller.Watches(
		&pxcv1.PerconaXtraDBCluster{},
		&handler.EnqueueRequestForObject{},
		builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
	)

```

controllers/databaseclusterbackup_controller.go
```diff
@@ -185,6 +180,32 @@ func (r *DatabaseClusterBackupReconciler) Reconcile(ctx context.Context, req ctr
...
+			databaseClusterNameLabel: backup.Spec.DBClusterName,
+			fmt.Sprintf(backupStorageNameLabelTmpl, backup.Spec.BackupStorageName): backupStorageLabelValue,
+		}
+		needUpdate = true
+	}
+	if len(backup.ObjectMeta.OwnerReferences) == 0 {
+		if err := controllerutil.SetControllerReference(cluster, backup, r.Client.Scheme()); err != nil {
+			return err
+		}
+		needUpdate = true
```
mayankshah1607, line 200: nitpick: I think we want to check the controller, as there can can multiple non-controlling owners (probably not in our case though, I'm just to be semantically correct..)
```suggestion
	if metav1.GetControllerOf(backup) == nil {
		if err := controllerutil.SetControllerReference(cluster, backup, r.Client.Scheme()); err != nil {
			return err
		}
		needUpdate = true
```

mayankshah1607 (approved):
LGTM, thanks!

I have some more non-crucial suggestions, but approving it to unblock :)

PR #440, opened 2024-06-27
EVEREST-859 | Ensure backups are deleted in foreground during clean-up

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-859

When uninstalling Everest via the CLI, the DB gets deleted before the backups can be cleaned-up properly

**Cause:**
We're not setting the `foregroundDeletion` finaliser.

**Solution:**
Set the finalizer at the time of performing the cleanup. Currently, the finalizers are added from different places (like from the API server). With this approach, we can unify the clean-up process, so that we don't have to explicitly add finalizers from the API server.

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databaseclusterbackup_controller.go
+++ b/controllers/databaseclusterbackup_controller.go
@@ -128,6 +128,14 @@ func (r *DatabaseClusterBackupReconciler) Reconcile(ctx context.Context, req ctr
 		}
 	}
 
+	// DBBackups are always deleted in foreground.
+	if controllerutil.AddFinalizer(backup, common.ForegroundDeletionFinalizer) {
+		if err := r.Update(ctx, backup); err != nil {
+			return reconcile.Result{}, err
+		}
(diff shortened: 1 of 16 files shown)


controllers/common/helper.go
```diff
@@ -659,13 +659,23 @@ func deleteBackupsForDatabase(
 			// Already deleting, continue to next.
 			continue
 		}
-		if err := c.Delete(ctx, &backup); err != nil {
+		if err := ForegroundDelete(ctx, &backup, c); err != nil {
```
michal-kralik, line 662: What if you delete the backup directly? Shouldn't it have the finalizer as well?

mayankshah1607: Updated

michal-kralik (approved):

oksana-grishchenko (approved):

PR #452, opened 2024-07-13
EVEREST-1247 | Re-design of operator upgrades

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1247

Re-design of operator upgrades

#### Why?

In 1.0.0 we added the ability to upgrade db operators. The initial design allowed upgrading each individual operator on the assumption that OLM creates an separate `InstallPlan` for each new operator. However this assumption is false and OLM instead aggregates all upgrades into a single InstallPlan. This required a small re-design of how Everest handles upgrades. In particular, we now upgrade all operators in a namespace at once.

The following changes are made:
- Remove the ability to upgrade a single databaseengine using the `everest.percona.com/upgrade-operator-to` annotation. This is due to the fact that we cannot actually control each individual operator upgrade, and we should remove this abstraction. Now the Everest server directly approves the InstallPlans listed in the dbengine status, and everest-operator only reconciles the status of the upgrade.
- Upgrade status is inferred using the CSVs.
- Everest has a namespace/dbenging locking mechanism that prevents any resource from being modified when an upgrade is taking place. However, in some rare cases, it is possible for the upgrade to go wrong and the lock staying in place indefinitely. With this PR, everest-operator will automatically remove the lock after 5mins


api/v1alpha1/databaseengine_types.go
```diff
@@ -89,15 +85,37 @@ type DatabaseEngineStatus struct {
...
+	if len(s.PendingOperatorUpgrades) == 1 {
+		return s.PendingOperatorUpgrades[0].TargetVersion
+	}
+	next := slices.MinFunc(s.PendingOperatorUpgrades, func(a, b OperatorUpgrade) int {
+		v1 := goversion.Must(goversion.NewVersion(a.TargetVersion))
+		v2 := goversion.Must(goversion.NewVersion(b.TargetVersion))
+		return v1.Core().Compare(v2.Core())
+	})
+	return next.TargetVersion
+}
```
oksana-grishchenko, line 105: How should Everest handle the patch releases? If there is a choice between `1.1.0` and `1.1.1` the function returns 1.1.0, does it mean the user suppose to go through each patch upgrade sequentially? It would be nice if Everest would suggest a single upgrade to the latest patch, but not sure if it's feasible. Anyway, let's at least add the the patch examples to the `api/v1alpha1/databaseengine_test.go`

mayankshah1607: Good point, I think we should skip to the latest patch.

> but not sure if it's feasible
Why not?

api/v1alpha1/databaseengine_types.go
```diff
@@ -51,16 +52,11 @@ const (
...
 
 const (
-	// DatabaseOperatorUpgradeAnnotation indicates that the database operator needs to be upgraded.
-	// The value of the annotation is the version to upgrade to.
-	DatabaseOperatorUpgradeAnnotation = "everest.percona.com/upgrade-operator-to"
 	// DatabaseOperatorUpgradeLockAnnotation is an annotation set on the database engine.
-	// If present and set to "true", Everest backend will prevent any resources from being modified in the namespace.
-	// This annotation is typically set by the Everest backend right before starting the operator upgrade.
+	// If present, the value should contain the timestamp of when the lock was set.
+	// Everest operator automatically removes this annotation (lock) after 5mins.
```
recharte, line 57: Are you sure 5 minutes is actually enough? What if the cluster is very slow at downloading images?

controllers/databaseengine_controller.go
```diff
@@ -103,41 +104,34 @@ func (r *DatabaseEngineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{},
 			r.Status().Update(ctx, dbEngine)
 	}
 
+	dbEngine.Status.OperatorVersion = version
+	if err := r.tryUnlockDBEngine(ctx, dbEngine); err != nil {
+		return ctrl.Result{}, err
+	}
```
recharte, line 131: Will this actually have the intended effect? If the annotation is set and nothing happens the reconciliation will never get triggered and the lock will stay locked. Should we somehow re-queue a reconciliation to the time when the unlocking should have already happened?

controllers/databaseengine_controller.go
```diff
@@ -181,135 +175,147 @@ func (r *DatabaseEngineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	lockedAt, err := time.Parse(time.RFC3339, lockedAtStr)
+	if err != nil {
+		return errors.Join(err, errors.New("cannot parse the value of 'everest.percona.com/upgrade-lock' annotation"))
+	}
+	// Check if it has exceeded the upgrade timeout.
+	if time.Now().After(lockedAt.Add(upgradeTimeout)) {
+		logger.Info("databaseengine upgrade timeout, force unlock triggered")
+		delete(annotations, everestv1alpha1.DatabaseOperatorUpgradeLockAnnotation)
+		dbEngine.SetAnnotations(annotations)
+		return r.Update(ctx, dbEngine)
```
recharte, line 278: Should we invert the condition and return nil to avoid duplicating the delete annotation logic found at the end of this func?
I'm not sure this is possible though, due to the rest of the conditions checked below.

controllers/databaseengine_controller.go
```diff
@@ -331,18 +337,60 @@ func (r *DatabaseEngineReconciler) listPendingOperatorUpgrades(
...
+		result = append(result, everestv1alpha1.OperatorUpgrade{
+			TargetVersion:  v,
+			InstallPlanRef: corev1.LocalObjectReference{Name: ipName},
+		})
+	}
 	return result, nil
 }
 
+// get the status and version of the operator deployment.
+// TODO: can we instead check the status of the corresponding CSV?
```
recharte, line 391: I'm not sure either :confused:

api/v1alpha1/databaseengine_types.go
mayankshah1607: 5mins was assuming most people run it on AWS and GKE, where I think 5mins would have been sufficient to download everything.. But I'm fine with a higher value too

controllers/databaseengine_controller.go
mayankshah1607: Hmm, fair point, although Kubernetes still does a period requeue regardless. But, still good to explicitly requeue to prevent unexpected delays

api/v1alpha1/databaseengine_types.go
recharte: I'm not sure what's the most appropriate value to be honest :confused:

controllers/databaseengine_controller.go
mayankshah1607: Done, I moved the unlock logic into a separate function variable to re-use it..

controllers/databaseengine_controller.go
mayankshah1607: I think this is possible, but then it will stop working with our Tilt setup because we don't use OLM there.
Once we have a Helm chart, I think we can simplify our tilt dev setup too..

controllers/databaseengine_controller.go
mayankshah1607: Done

api/v1alpha1/databaseengine_types.go
mayankshah1607: Okay, let's change it to 10mins, to be on the safer side :)

recharte (approved):

api/v1alpha1/databaseengine_types.go
oksana-grishchenko: no particular reason in mind, I just wasn't sure if the upgrade is going to work well with all of the components (catalog, VS) but seems like it should.

oksana-grishchenko (approved):

PR #462, opened 2024-07-23
EVEREST-1244 PXC restoration status fix

oksana-grishchenko:
**PXC restoration status update**
---
**Problem:**
EVEREST-1244

**Problem:**
When a PXC cluster is in the "Stopping Cluster" or "Starting Cluster" state during the restoration it's not considered as a restoring status and as a result the FE doesn't block actions over the cluster. 

This behaviour was intentionally introduced in https://github.com/percona/everest-operator/pull/435, because
>  in our reconciliation, we depend on explicitly checking the states Starting Cluster and Stopping Cluster, to handle pausing/unpausing the DB when restoring to a new cluster. Since these states disappeared (into Running/Starting), this logic started to fail. 

**Solution:**
1. revert the https://github.com/percona/everest-operator/pull/435, so Everest would again merge the states such as `Starting Cluster` under `Running`, and `Stopping Cluster` under `Starting`.
2. handle the pxc cluster pausing logic based on the `pxc-restore` CR status but not on the `dbr` status 

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/providers/pxc/provider.go
+++ b/controllers/providers/pxc/provider.go
@@ -19,6 +19,7 @@ package pxc
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 
 	goversion "github.com/hashicorp/go-version"
@@ -110,7 +111,7 @@ func New(
 	if err := p.ensureDefaults(ctx); err != nil {
 		return nil, err
 	}
-	if err := p.handlePauseForPXCRestore(ctx); err != nil {
+	if err := p.handlePauseForPXCRestore(ctx, opts); err != nil {
 		return nil, err
 	}
 	return p, nil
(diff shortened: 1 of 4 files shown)


controllers/providers/pxc/provider.go
```diff
@@ -130,23 +131,27 @@ func (p *Provider) Apply(ctx context.Context) everestv1alpha1.Applier {
 // The pause may be overwritten since the Everest operator will also fight for the desired pause state
 // set by the user.
 // To avoid this race condition, we will explicitly inspect every phase of the restore, and set pause accordingly.
-func (p *Provider) handlePauseForPXCRestore(ctx context.Context) error {
-	restores, err := common.ListDatabaseClusterRestores(ctx, p.C, p.DB.GetName(), p.DB.GetNamespace())
-	if err != nil {
-		return err
+func (p *Provider) handlePauseForPXCRestore(ctx context.Context, opts providers.ProviderOptions) error {
+	restores := &pxcv1.PerconaXtraDBClusterRestoreList{}
```
mayankshah1607, line 135: Since this is handled in the db-cluster controller, shouldn't it also watch pxc-restores now? Otherwise there could be a delay in this reconciliation..

mayankshah1607:
This approach looks fine to me, but I still have one question:

> as a result the FE doesn't block actions over the cluster.

Can't the FE consider these Start/Stop states as well to disable the UI actions? Just like it does for the Restoring state?

oksana-grishchenko:
> Can't the FE consider these Start/Stop states as well to disable the UI actions? Just like it does for the Restoring state?

@mayankshah1607 my concern is that the same statuses are used for the regular pause/unpause (not during the restoration) and then the UI needs to distinguish those states somehow

controllers/providers/pxc/provider.go
oksana-grishchenko: A good point. The db-cluster controller watches `DatabaseClusterRestore`, and `DatabaseClusterRestore` should be reconciled each time the upstream restore object is reconciled... need to take a closer look to it

michal-kralik (approved):

controllers/providers/pxc/provider.go
oksana-grishchenko: Checked it. There could be a delay indeed if the restoration doesn't go the happy path, so in that case we need to watch the `pxc-restore` resources to be sure the db gets reconciled when the `pxcv1.RestoreStopCluster` status is met. 
The downside is that for pxc we're going to have extra reconciliation calls, bc we watch both `dbr` and `pxc-restore`. I don't really like the solution, but it at least it solves the problems we faced. @mayankshah1607 wdyt?

recharte (approved):

controllers/providers/pxc/provider.go
oksana-grishchenko: Resolving the thread bc
- the suggestion was implemented
- Mayank is temporary out of work
- need to move forward with the release

PR #465, opened 2024-07-26
EVEREST-1278 Fix new PG DBs getting stuck in Unknown state after deleting a PG DB

recharte:
**Fix new PG DBs getting stuck in Unknown state after deleting a PG DB**
---
**Problem:**
EVEREST-1278

New PG DBs get stuck in the Unknown state after deleting a PG DB

**Related pull requests**

- https://github.com/percona/percona-postgresql-operator/pull/848

**Cause:**
Due to a bug in PGO v2.4.0 the PGO reconciliation loop is getting stuck after the deletion of a DB which causes any new PG DBs provisioned in Everest to be stuck in the Unknown state.

**Solution:**
Until [K8SPG-616](https://perconadev.atlassian.net/browse/K8SPG-616) is released we work around the issue by restarting the PG operator deployment after deleting a DB.

**CHECKLIST**
---
**Jira**

**Tests**

[K8SPG-616]: https://perconadev.atlassian.net/browse/K8SPG-616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/controllers/common/helper.go
+++ b/controllers/common/helper.go
@@ -621,7 +621,7 @@ func GetBackupStorageIndexInPGBackrestRepo(
 	return -1
 }
 
-// HandleUpstreamClusterCleanup handles the cleanup of the psdmb objects.
+// HandleUpstreamClusterCleanup handles the cleanup of the upstream DB objects.
 // Returns true if cleanup is complete.
 func HandleUpstreamClusterCleanup(
 	ctx context.Context,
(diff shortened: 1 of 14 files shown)


controllers/common/helper.go
```diff
@@ -694,3 +696,32 @@ func GetRecommendedCRVersion(
...
+	err := c.Get(ctx, name, deployment)
+	if err != nil {
+		return err
+	}
+	if deployment.Spec.Template.ObjectMeta.Annotations == nil {
+		deployment.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
+	deployment.Spec.Template.ObjectMeta.Annotations["kubectl.kubernetes.io/restartedAt"] = metav1.Now().Format(time.RFC3339)
+	return c.Update(ctx, deployment)
+}
```
mayankshah1607, line 727: This annotation needs to come in through the CSV, not the Deployment, otherwise the CSV can override the Deployment annotations before the deployment-controller has had a chance to observe the updated annotation.

This is what we do in the Everest server, I already faced this problem once 😅 - https://github.com/percona/everest/blob/02f5f93e33b9ab9803d01bbb7715869fc4fbfdef/pkg/kubernetes/kubernetes.go#L838

mayankshah1607: But then our dev setup will be broken since we don't use OLM there 🤷🏻

oksana-grishchenko (commented):
UPD: During the DS it was decided to not include the PG and the PSMDB operators upgrades to the Everest 1.1 and wait for the fixes on the operator's side. 
 
So we can keep this workaround PR as draft just in case the PG 2.4.1 is delayed to the time we want Everest 1.2.

recharte:
This PR was opened to attempt to workaround an issue with PGO v2.4.0. Because of this issue, we decided not to release everest with PGO v2.4.0. Now that we are integrating with PGO v2.4.1 this issue is no longer present, hence I’m closing this PR.

PR #468, opened 2024-07-29
EVEREST-1275 Fix PG repos reconciliation

oksana-grishchenko:
**Fix PG repos reconciliation**
---
**Problem:**
EVEREST-1275

Repos in the pg cluster configuration got messed up in the described below case. The repos misconfiguration led to the backups deletion which is why the ticket was created.  

_The case:_
   1. there is a backup without schedule (it may happen bc the schedule was deleted or if an on-demand backup was taken) pointing to `storageA`.  (in PG: `repo2` pointing to `storageA`, no schedule)
   2. add a schedule pointing to `storageB`. 
         Expected: 
                     - `repo2` pointing to `storageA`, no schedule
                     - `repo3` pointing to `storageB`, with schedule
         Actual: 
                      - `repo2` pointing to `storageB`, with schedule
                      - `repo3` pointing to `storageA`, no schedule

**Cause:**
Everest was adding new schedules to repos before it has checked the repos of the existing backups. 

**Solution:**
- Refactor the method which reconciles pg repos to make it observable and be able to find the bug.
- The fix: swap the order of the two processes (adding new schedules and checking the existing backups repos). This is the only change in the existing logic.  
- Add a unit test to check the case does not fail anymore. 

**CHECKLIST**
---
**Jira**

**Tests**


controllers/providers/pg/applier.go
```diff
@@ -1043,346 +1039,46 @@ func reconcilePGBackRestRepos(
...
-		err = addBackupStorageCredentialsToPGBackrestSecretIni(sType, pgBackRestSecretIni, repo.Name, backupStoragesSecrets[backupRequest.Spec.BackupStorageName])
-		if err != nil {
-			return []crunchyv1beta1.PGBackRestRepo{},
-				map[string]string{},
-				[]byte{},
-				errors.Join(err, errors.New("failed to add backup storage credentials to PGBackrest secret data"))
-		}
+	err = reposReconciler.reconcileBackups(backups, backupStorages, backupStoragesSecrets, db)
+	if err != nil {
+		return []crunchyv1beta1.PGBackRestRepo{}, map[string]string{}, []byte{}, err
```
oksana-grishchenko, line 1062: besides the refactoring, the only change is that now we call `reconcileBackups` before `addNewSchedules` but not after it.

michal-kralik (approved):

PR #474, opened 2024-08-02
EVEREST-1287 Change affinity default settings

oksana-grishchenko:
**Affinity default settings**
---
**Problem:**
EVEREST-1287

Affinity was set to `none` by default. Only for EKS it was set to `kubernetes.io/hostname`

Traced back the git history - this condition for EKS was implemented in January 2023, but the reasoning wasn't really clear. 
Checked it with @gen1us2k, who fortunately can remember the reasons:
- EKS was the main focus 
- The setting caused the db clusters never came up in a single node k8s cluster.  

Not sure, what was the reason which cased the db clusters never came up back then, but checked it now - all 3 types can come up with the affinity settings set by default in a single-node kind cluster. 

So the condition was deleted and the setting is applied by default. 

Any thoughts/feedback is appreciated. 

**CHECKLIST**
---
**Jira**

**Tests**


mayankshah1607:
I may be missing something here, but, by setting the affinity topology key to `kubernetes.io/hostname`, aren't we asking the scheduler to schedule all pods on the same node? Which seems like the opposite of what the original ticket describes - having to schedule pods across different availability zones 🤔

oksana-grishchenko:
@mayankshah1607 thing that might be confusing here is that the `Affinity` key actually refers to `AntiAffinity` under the hood ([PXC](https://github.com/percona/percona-xtradb-cluster-operator/blob/f76e86f2ebfa18fef11d508a4d59d7edc7141193/pkg/apis/pxc/v1/pxc_types.go#L514), [PSMDB](https://github.com/percona/percona-server-mongodb-operator/blob/f10e1aa6286cb313bee882e6a25938e2b9106b58/pkg/apis/psmdb/v1/psmdb_types.go#L410), [PG](https://github.com/percona/everest-operator/blob/35113c70e73461b6aa31e3c03c422198ca0f2124/controllers/providers/pg/provider.go#L41))

As I understood, setting `AntiAffinity` to `kubernetes.io/hostname` configures the scheduler to not set pods to the same node within the same availability zone. As the ticket says, the problem is 
> if that node goes down the whole DB cluster can go down

so the idea is at least to schedule pods to different nodes by default. 

If the user wants further improvements like spreading across different zones, they should be able to configure the topology to something else, e.g. `kubernetes.io/zone`, but it's out of scope for this ticket.

mayankshah1607:
> Affinity key actually refers to AntiAffinity under the hood

Okay, I see. Then it makes sense 😅 Thanks for explaining!

mayankshah1607 (approved):

michal-kralik (approved):

PR #477, opened 2024-08-08
EVEREST-1316 Backups repos reconciliation fix

oksana-grishchenko:
**Backups repos reconciliation fix**
---
**Problem:**
EVEREST-1316

Changes:

1. For backups and new schedules: first check if there is any existing repo which can be reused and only if there is none - create a new repo
2. When creating an on-demand backup, rely on the repo name but not on it's index - this approach is more reliable. 
3. Sort the repos by name. When schedules are deleted, the pg cluster might get into the state when it has [repo1, repo3, repo2], so the repos can switch places although they still pointing out to the same bucket as before. It's confusing, so from now on we always sort repos by name
4. Added tests to cover the failing cases.  

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/common/helper.go
+++ b/controllers/common/helper.go
@@ -601,8 +601,8 @@ func IsDatabaseClusterRestoreRunning(
 	return false, nil
 }
 
-// GetBackupStorageNameInPGBackrestRepo returns the name of the backup storage in the pgbackrest repo list.
-func GetBackupStorageNameInPGBackrestRepo(
+// GetRepoNameByBackupStorage returns the name of the repo that corresponds to the given backup storage
+func GetRepoNameByBackupStorage(
 	backupStorage *everestv1alpha1.BackupStorage,
 	repos []crunchyv1beta1.PGBackRestRepo,
 ) string {
--- a/controllers/common/helper.go
+++ b/controllers/common/helper.go
@@ -601,24 +601,24 @@ func IsDatabaseClusterRestoreRunning(
 	return false, nil
 }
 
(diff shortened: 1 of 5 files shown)


controllers/common/helper.go
```diff
@@ -601,24 +601,24 @@ func IsDatabaseClusterRestoreRunning(
 	return false, nil
 }
 
-// GetBackupStorageIndexInPGBackrestRepo returns the index of the backup storage in the pgbackrest repo list.
-func GetBackupStorageIndexInPGBackrestRepo(
+// GetBackupStorageNameInPGBackrestRepo returns the name of the backup storage in the pgbackrest repo list.
+func GetBackupStorageNameInPGBackrestRepo(
```
recharte, line 605: Just a nitpick but we are actually returning the name of the repo not the name of the backup storage, right? If that's the case we should consider renaming the function to avoid confusion with the name of the backup storage CR.

recharte (approved):
As discussed offline, I hate how complex this fix had to be because of the existing code base for this repos reconciliation but I believe that the logic is sound.
Nonetheless, we must refactor this code to be more maintainable, in the current state it's terrible... Opened [EVEREST-1319](https://perconadev.atlassian.net/browse/EVEREST-1319) for this purpose.

controllers/common/helper.go
oksana-grishchenko: Right, thanks for noticing this. Renamed to `GetRepoNameByBackupStorage`

PR #495, opened 2024-08-22
EVEREST-1305 Operator sharding API

oksana-grishchenko:
**Mongo sharding**
---
**Problem:**
EVEREST-1305

**What's new**
- Mongo sharding: Enabling, configure number of shards and configservers. 
- Proxy: for non-sharded cluster expose replset[0] directly as Everest did before, for sharded clusters use .mongos to expose cluster. 

**Related pull requests**

- https://github.com/percona/everest/pull/615

**CHECKLIST**
---
**Jira**

**Tests**


api/v1alpha1/databasecluster_types.go
```diff
@@ -310,6 +311,14 @@ type Monitoring struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 }
 
+// Sharding are the sharding options. Available only for psmdb
+type Sharding struct {
+	Enabled      bool                 `json:"enabled"`
+	Shards       int32                `json:"shards"`
+	ConfigServer *psmdbv1.ReplsetSpec `json:"configServer,omitempty"`
+	Proxy        *psmdbv1.MongosSpec  `json:"proxy,omitempty"`
```
oksana-grishchenko, line 319: Concerned about exposing the psmdb structs, should we create Everests's own abstraction with only the fields we're going to configure and add the new fields once requested?

api/v1alpha1/databasecluster_types.go
```diff
@@ -310,6 +311,14 @@ type Monitoring struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 }
 
+// Sharding are the sharding options. Available only for psmdb
+type Sharding struct {
+	Enabled      bool                 `json:"enabled"`
+	Shards       int32                `json:"shards"`
```
oksana-grishchenko, line 317: The number will be mapped to the psmdb `.Spec.Replsets` section so that a cluster will have as many replsets as the number of shards `N`. 
The list of the replsets by default contains only one replset `rs0`, but with the sharding is enabled the list will contain N (probably identical) replsets `rs0`, `rs1`, ... `rs(N-1)`

api/v1alpha1/databasecluster_types.go
recharte: Why do we need proxy here? We already have a proxy in `.spec.proxy`

api/v1alpha1/databasecluster_types.go
```diff
@@ -310,6 +311,14 @@ type Monitoring struct {
...
+	ConfigServer *psmdbv1.ReplsetSpec `json:"configServer,omitempty"`
```
recharte, line 318: Yeah maybe we shouldn't expose the psmdbv1.ReplsetSpec. I think that right now we only need to actually change the number of config server replicas so maybe we create a
```
type ConfigServer struct {
}
```
and we expand it in the future if needed. WDYT?

api/v1alpha1/databasecluster_types.go
oksana-grishchenko: It's actually the `mongos` section of the sharding, which accroding to @hors is basically proxy. Mongo doesn't have any other proxies. But it's a good question if we can reuse the existing `.spec.proxy` section and just map it to the `.sharding.mongos` under the hood, let me check it.

recharte: That's how we were doing it in the very beginning of Everest. For some random reason we used to always deploy a sharded cluster with a single replicaset

oksana-grishchenko: Yes, we can definitely reuse it, we even have this `mongos` type 👍

api/v1alpha1/databasecluster_types.go
oksana-grishchenko: Sounds good to me, especially considering that it's the only struct we need to expose.

oksana-grishchenko: Perhaps we don't even need a new object for configServer. We might use `.db.Spec.Proxy.Replicas` as we do for `pxc` and `pg`. @recharte  wdyt?

recharte: No, ConfigServer is different from Mongos, `.db.Spec.Proxy.Replicas` should match up to `.spec.sharding.mongos.size`

mayankshah1607 (approved):

PR #500, opened 2024-08-28
EVEREST-1264 | Fix Secret reconciliation for MonitoringConfig controller

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1264

MonitoringConfigs stopped working after https://github.com/percona/everest-operator/pull/463 was merged.

**Cause:**
https://github.com/percona/everest-operator/pull/463 changes the way how MonitoringConfigs are referenced by the DB. Earlier, MonitoringConfigs and its Secret were created in the same namespace as the VMAgent, i.e, `everest-monitoring.` With https://github.com/percona/everest-operator/pull/463, MonitoringConfigs and its Secret are now created in the DB namespace, and VMAgent can no longer reference the secret since it belongs to a different namespace.

**Solution:**

- Add additional logic to copy the Secret to the `everest-monitoring` namespace
- Secret is automatically cleaned up with the help of a finalizer

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/monitoringconfig_controller.go
+++ b/controllers/monitoringconfig_controller.go
@@ -348,6 +348,12 @@ func (r *MonitoringConfigReconciler) cleanupSecrets(ctx context.Context, mc *eve
 // reconcileSecret copies the source MonitoringConfig secret onto the monitoring namespace.
 // Returns the name of the newly created/updated secret.
 func (r *MonitoringConfigReconciler) reconcileSecret(ctx context.Context, mc *everestv1alpha1.MonitoringConfig) (string, error) {
+	// If the MonitoringConfig is in the monitoring namespace, we don't have to do any additional work.
+	// Use the secret as-is.
+	if mc.GetNamespace() == r.monitoringNamespace {
+		return mc.Spec.CredentialsSecretName, nil
+	}
+
 	// Get the secret in the MonitoringConfig namespace.
 	secret := &corev1.Secret{}
 	err := r.Get(ctx, types.NamespacedName{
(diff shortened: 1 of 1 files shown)


michal-kralik (approved):

controllers/monitoringconfig_controller.go
```diff
@@ -305,3 +318,78 @@ func deduplicateRemoteWrites(remoteWrites []interface{}) []interface{} {
...
+		labels[monitoringConfigRefNamespaceLabel] = mc.GetNamespace()
+		dst.SetLabels(labels)
+		return nil
+	}); err != nil {
+		return "", err
+	}
+	// Add a clean-up finalizer in the parent MonitoringConfig.
+	if controllerutil.AddFinalizer(mc, monitoringConfigSecretCleanupFinalizer) {
+		return dstSecretName, r.Update(ctx, mc)
+	}
```
oksana-grishchenko, line 393: should this method check `mc.GetDeletionTimestamp().IsZero()` before adding the finalizer? I mean, is there a chance it will be added again here after it was deleted?

mayankshah1607: I don't think so..

[Here](https://github.com/percona/everest-operator/pull/500/files#diff-1d6b7319d3daf756d50f49cb113d91450fbe3022146f58e9793a931a8f4e01a0R84) we're already checking for the deletion timestamp at the top level `Reconcile`. If it is not zero, we cleanup and return early, even before `reconcileSecrets` gets called. So I don't think the finalizer will be added again

PR #510, opened 2024-09-05
EVEREST-1415 Deleting status fix

oksana-grishchenko:
**Deleting status fix**
---
**Problem:**
EVEREST-1415

Everest cluster does not follow the further upstream cluster statuses after deletion. As a result if the upstream cluster deletion takes long (for example, a sharded psmdb cluster with 5 nodes), the psmdb cluster hangs in the “stopping“ state but in Everest there is no DatabaseCluster shown anymore as if it is deleted. 

**Cause:**
Everest has this `everest.percona.com/upstream-cluster-cleanup` finalizer which ensures that the upstream cluster deletion is triggered and after this deletion Everest removes the finalizer  which unblocks the DatabaseCluster CR deletion. So the DatabaseCluster CR is deleted but the upstream psmdb still exists in the “stopping“ state. 

**Solution:**
Do not remove the `everest.percona.com/upstream-cluster-cleanup` finalizer until the upstream cluster is indeed deleted. 

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/common/helper.go
+++ b/controllers/common/helper.go
@@ -665,8 +665,6 @@ func HandleUpstreamClusterCleanup(
 			}
 			return false, err
 		}
-		controllerutil.RemoveFinalizer(database, UpstreamClusterCleanupFinalizer)
-		return true, c.Update(ctx, database)
 	}
 	return true, nil
 }
--- a/e2e-tests/tests/upgrade/pg/90-delete-clusters.yaml
+++ b/e2e-tests/tests/upgrade/pg/90-delete-clusters.yaml
@@ -1,6 +1,6 @@
 apiVersion: kuttl.dev/v1
 kind: TestStep
 commands:
-  - command: kubectl -n $NAMESPACE delete db test-pg-cluster
   - command: kubectl patch postgresclusters test-pg-cluster -n $NAMESPACE -p '{"metadata":{"finalizers":null}}' --type merge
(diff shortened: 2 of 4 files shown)


controllers/common/helper.go
```diff
@@ -665,8 +665,6 @@ func HandleUpstreamClusterCleanup(
 			}
 			return false, err
 		}
```
mayankshah1607, line 667: LGTM, but one suggestion that's unrelated to the changes in this PR. 'm wondering if the code above this can be simplified:

I see there's a Get and then Delete. The Get seems redundant. Why not just do a Delete? If success, then return false (retry). If NotFoundErr, then remove finalizer and update

mayankshah1607 (approved):

michal-kralik (approved):

controllers/common/helper.go
oksana-grishchenko: @mayankshah1607 I think the Get is there bc of how the `common.HandleUpstreamClusterCleanup` is being called from the provider: 
https://github.com/percona/everest-operator/blob/cbfab94d6cff424e51c98561af53bff749209d97/controllers/providers/psmdb/provider.go#L170
So provider receives the technology-specific empty object and then gets the actual object from k8s to delete it. Otherwise the logic that gets the upstream CR would need to be duplicated in each `Cleanup()` implementation.

PR #511, opened 2024-09-06
EVEREST-1424 | Apply new affinity settings for new clusters only

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-0

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**


controllers/providers/psmdb/provider.go
```diff
@@ -46,6 +46,9 @@ type Provider struct {
 	*psmdbv1.PerconaServerMongoDB
 	providers.ProviderOptions
 
+	// currentPSMDB holds the current PXC spec.
```
tplavcic, line 49:
```suggestion
	// currentPSMDB holds the current PSMDB spec.
```

controllers/providers/pg/provider.go
```diff
@@ -41,6 +41,8 @@ type Provider struct {
 	*pgv2.PerconaPGCluster
 	providers.ProviderOptions
 	clusterType common.ClusterType
+	// currentPGSpec holds the current PXC spec.
```
tplavcic, line 44:
```suggestion
	// currentPGSpec holds the current PG spec.
```

tplavcic (changes requested):

michal-kralik (approved):

oksana-grishchenko (approved):

controllers/providers/pxc/applier.go
```diff
@@ -95,6 +95,12 @@ func (p *applier) Engine() error {
 		pxc.Spec.PXC.PodSpec.Resources.Limits[corev1.ResourceMemory] = p.DB.Spec.Engine.Resources.Memory
 		pxc.Spec.PXC.PodSpec.Resources.Requests[corev1.ResourceMemory] = p.DB.Spec.Engine.Resources.Memory
 	}
+	// We preserve the settings for existing DBs, otherwise restarts are seen when upgrading Everest.
+	// TODO: Remove this once we figure out how to apply such spec changes without automatic restarts.
+	// See: https://perconadev.atlassian.net/browse/EVEREST-1413
+	if p.DB.Status.Status == everestv1alpha1.AppStateReady {
+		pxc.Spec.PXC.PodSpec.Resources = p.currentPerconaXtraDBClusterSpec.PXC.PodSpec.Resources
```
recharte, line 102: @mayankshah1607 Doesn't this mean that users won't be able to change the resources of a DB that's in the Ready state? With this change we are breaking the feature of vertical scaling, right?

mayankshah1607: @recharte you're right...kinda.

I didn't think of this earlier, that by preserving the settings as-is when the state is "running", a user cannot apply his desired changes anymore.

So I tested it. And turns out, that it still works, unexpectedly. I was able to vertically scale the cluster even though the code here preserves the original settings when the state is running. Reason is that when you vertically scale, it also changes the proxy pod resources. And as a result, the DB goes into the "initialising" state, and the new settings for the pxc pod also get applied (as a side-effect).

Basically, it still "accidentally" works.. This still feels wrong to me, and I'd like it to work based on expectations and not surprises. So I created a proper fix for it: https://github.com/percona/everest-operator/pull/513

recharte: > it also changes the proxy pod resources
I think this might only true for PXC and not for the other engines

PR #513, opened 2024-09-10
EVEREST-1424 | Cannot apply PXC resource settings after adding restart protection mechanism

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1424

Cannot update PXC resource settings after adding the restart protection mechanism in https://perconadev.atlassian.net/browse/EVEREST-1424

- https://perconadev.atlassian.net/browse/EVEREST-1424

For more details, see comment https://github.com/percona/everest-operator/pull/511#discussion_r1751444476

**Cause:**

In the above mentioned PR, the resource settings for PXC are preserved if the DB is in running state. This means, that while there will not be any restarts when ugprading to 1.2.0, a user also cannot update these settings because the DB is running, and the older settings are preserved/overwritten.

**Solution:**

Add an additional check to detect when the DB spec may have potentially changed. We do this by tracking `.metadata.Generation`

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/api/v1alpha1/databasecluster_types.go
+++ b/api/v1alpha1/databasecluster_types.go
@@ -356,6 +356,8 @@ type DatabaseClusterSpec struct {
 
 // DatabaseClusterStatus defines the observed state of DatabaseCluster.
 type DatabaseClusterStatus struct {
+	// ObservedGeneration is the most recent generation observed for this DatabaseCluster.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 	// Status is the status of the cluster
 	Status AppState `json:"status,omitempty"`
 	// Hostname is the hostname where the cluster can be reached
--- a/controllers/providers/pxc/applier.go
+++ b/controllers/providers/pxc/applier.go
@@ -95,10 +95,13 @@ func (p *applier) Engine() error {
 		pxc.Spec.PXC.PodSpec.Resources.Limits[corev1.ResourceMemory] = p.DB.Spec.Engine.Resources.Memory
 		pxc.Spec.PXC.PodSpec.Resources.Requests[corev1.ResourceMemory] = p.DB.Spec.Engine.Resources.Memory
 	}
+	hasDBSpecChanged := func() bool {
+		return p.DB.Status.ObservedGeneration > 0 && p.DB.Status.ObservedGeneration != p.DB.Generation
+	}
 	// We preserve the settings for existing DBs, otherwise restarts are seen when upgrading Everest.
 	// TODO: Remove this once we figure out how to apply such spec changes without automatic restarts.
 	// See: https://perconadev.atlassian.net/browse/EVEREST-1413
-	if p.DB.Status.Status == everestv1alpha1.AppStateReady {
+	if p.DB.Status.Status == everestv1alpha1.AppStateReady && !hasDBSpecChanged() {
 		pxc.Spec.PXC.PodSpec.Resources = p.currentPerconaXtraDBClusterSpec.PXC.PodSpec.Resources
 	}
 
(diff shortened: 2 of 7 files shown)


controllers/providers/pxc/applier.go
```diff
@@ -116,10 +116,14 @@ func (p *applier) Engine() error {
...
 	}
+	hasDBSpecChanged := func() bool {
+		return p.DB.Status.ObservedGeneration > 0 && p.DB.Status.ObservedGeneration != p.DB.Generation
+	}
 	// We preserve the settings for existing DBs, otherwise restarts are seen when upgrading Everest.
+	// Additionally, we also need to check for the spec changes, otherwise the user can never voluntarily change the resource setting.
 	// TODO: Remove this once we figure out how to apply such spec changes without automatic restarts.
 	// See: https://perconadev.atlassian.net/browse/EVEREST-1413
-	if p.DB.Status.Status == everestv1alpha1.AppStateReady {
+	if p.DB.Status.Status == everestv1alpha1.AppStateReady && !hasDBSpecChanged() {
```
oksana-grishchenko, line 126: It should work fine since operator always updates status as a last step of reconciliation, so LGTM.  

should we apply the check for all engines affinity and also for the pxc resources? Basically, it means everywhere we have this
```
// TODO: Remove this once we figure out how to apply such spec changes without automatic restarts.
```

mayankshah1607: > should we apply the check for all engines affinity
Affinity settings cannot be updated by the user yet, so I think it's fine to skip it. The reason we applied it for PXC resource settings is because it can be changed by the user, so we need to be able to detect spec changes.. So, for affinity, I don't think it matters. Moreover, I'd eventually like to get rid of all these manual if-else checks and implement a better mechanism, which takes the current and previous Everest version into account..

> also for the pxc resources?
Is there any other pxc resources you're referring to other than the one fixed in this PR?

oksana-grishchenko: @mayankshah1607 I mean, I was confused bc the check is applied to the affinity setting now, look at the next line:
```
 pxc.Spec.PXC.PodSpec.Affinity = p.currentPerconaXtraDBClusterSpec.PXC.Affinity
```
it should've been applied to the line 102 instead. Then I would've no more questions :)

api/v1alpha1/databasecluster_types.go
```diff
@@ -356,6 +356,8 @@ type DatabaseClusterSpec struct {
 
 // DatabaseClusterStatus defines the observed state of DatabaseCluster.
 type DatabaseClusterStatus struct {
+	// ObservedGeneration is the most recent generation observed for this DatabaseCluster.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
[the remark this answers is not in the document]
recharte: I know that other controllers also use this mechanism so maybe it's fine...

controllers/providers/pxc/applier.go
mayankshah1607: Ohh crap. I messed up something when refactoring, thanks for pointing out..
Yeah, it should have been for the resources only 🙈

oksana-grishchenko (approved):

api/v1alpha1/databasecluster_types.go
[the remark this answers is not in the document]
mayankshah1607: Yeah, this is a pretty common pattern in operators for change detection..

I understand that this field has little value to the users, but I also don't see a better/simpler way to do this. We can always keep the API definition from the CRD (and skip this field).. not sure how feasible though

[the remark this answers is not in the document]
recharte: Yeah, I think it's okay :slightly_smiling_face: 

> We can always keep the API definition different from the CRD (and skip this field).. not sure how feasible though

No need to do that

controllers/providers/pxc/applier.go
```diff
@@ -117,6 +120,7 @@ func (p *applier) Engine() error {
 		Advanced: common.DefaultAffinitySettings().DeepCopy(),
 	}
 	// We preserve the settings for existing DBs, otherwise restarts are seen when upgrading Everest.
+	// Additionally, we also need to check for the spec changes, otherwise the user can never voluntarily change the resource setting.
```
recharte, line 123: You added this comment here but the spec change check was added above :sweat_smile:

mayankshah1607: My bad. Left overs from some refactoring. Fixed 😅

recharte (approved):

PR #534, opened 2024-10-07
EVEREST-1511 | Allow reconciling objects based on namespace labels

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1511

To support installation via Helm chart, we would like to decouple the installation process of the core Everest components and the DB namespaces. This means, first the Operator, API server, OLM and Monitoring components are installed, and then the DB namespaces.

However, when the Everest operator is deployed before the DB namespaces, it goes into a crashed state, and as a result, the first step on the Helm installation never succeeds.

**Cause:**

Currently, Everest operator requires that we set the `DB_NAMESPACES` environment variable during bootstrap. It reads this variable, setups up caches for those namespaces and checks if those namespaces actually exist. If they don't, it panics. This cannot work with the Helm chart requirements, as the DB namespaces will be created afterwards, and it cannot always know the `DB_NAMESPACES` before hand.

Moreover, patching the DB_NAMESPACES in the Deployment/Subscription manifest in Helm gets very messy.

**Solution:**

Support reconciling objects based on labels:
- We already set the `app.kubernetes.io/managed-by=everest` label on DB namespaces. With this PR, we can skip specifying `DB_NAMESPACE`, and the operator will reconcile objects only from those namespaces which have this label. Since controller-runtime does not allow filtering objects based on namespace labels, we cache all namespaces and filter out the reconcile requests. This can cause some extra memory consumption, but should not be a concern.
- This label may be configured using the `--namespace-label-filter` flag. Multiple labels may be set and they will be AND-ed.
- DB_NAMESPACES environment variable may still be specified, and that will take precedence over the label filter. This is done to ensure backward compatibility and compatibility with the CLI
- Added some code cleanup and refactoring as well.

**CHECKLIST**
---
**Jira**

**Tests**


controllers/databasecluster_controller.go
```diff
@@ -691,6 +689,7 @@ func (r *DatabaseClusterReconciler) initWatchers(controller *builder.Builder) {
 	if r.Scheme.Recognizes(common.PSMDBGVK) {
 		controller.Watches(
 			&psmdbv1.PerconaServerMongoDB{},
+
```
recharte, line 692:
```suggestion
```

recharte (approved):
Changes look good.
Soon we'll need to migrate away from the initial cache configuration approach because maintaining both is messy.

mayankshah1607:
Created a ticket for the DB_NAMESPACES removal: https://perconadev.atlassian.net/browse/EVEREST-1553

PR #547, opened 2024-10-09
EVEREST-1511 | Restart operator pod if new DB CRDs are installed

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
https://github.com/percona/everest-operator/pull/534 provides a way to dynamically watch namespaces based on labels, rather than setting the `DB_NAMESPACES` var. However, using this mechanism causes the DB reconciliation to get stuck.

**Cause:**

When bootstrapping the various controllers, we add watches to the DB CRs only if the CRDs exist first. This step happens only once, before the controller is started. With https://github.com/percona/everest-operator/pull/534, we're now allowed to start the operator without the DB namespaces. This means that when the controller starts, none of the DB CRDs exist, and as a result, the controllers are never watching the DB CRs.

This did not happen earlier because when you update the `DB_NAMESPACES` var, it leads to  a restart anyway, and hence all the controllers get setup again with the new watchers.

**Solution:**

This PR adds some code to inspect the DB engines and see if their corresponding CRDs are registered to the scheme. If it finds a DB engine whose scheme is not registered, then it restarts its own pod.

This is not the most elegant solution, but the simplest one with the least amount of code. A better solution would be to create a parent controller for DB, DBC and DBB, which watches for DBEngines, and dynamically creates a controller for each engine type. But this would significantly increase the scope of this work, so will this option out for now.

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/controllers/databaseengine_controller.go
+++ b/controllers/databaseengine_controller.go
@@ -211,7 +211,6 @@ func (r *DatabaseEngineReconciler) restartIfNeeded(ctx context.Context) (bool, e
 	if err := r.List(ctx, dbEngines); err != nil {
 		return false, fmt.Errorf("failed to list DatabaseEngines: %w", err)
 	}
-	unregistered := 0
 	for _, dbEngine := range dbEngines.Items {
 		// Wait until all DB engines are either installed/not installed.
 		// This way we can avoid redundant restarts.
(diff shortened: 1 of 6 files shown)


controllers/databaseengine_controller.go
```diff
@@ -177,11 +189,60 @@ func (r *DatabaseEngineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+		group, found := operatorEngineTypeToCRDGroup[dbEngine.Spec.Type]
+		if !found {
+			return false, fmt.Errorf("unknown engine type '%s'", dbEngine.Spec.Type)
+		}
+		// Ideally we would also like to check if all registered controllers are also watching the CRs,
+		// but since that's tricky to accomplish, we will only check if the CRDs are registered to the scheme,
+		// since we typically perform that step along with configuring the watches.
+		if !r.Scheme.IsGroupRegistered(group) {
+			unregistered++
```
recharte, line 234: Why don't we just return here instead of counting?

mayankshah1607: Fixed

recharte (approved):

PR #565, opened 2024-10-29
EVEREST-1599 add PGO 2.5.0

recharte:
**Add PGO 2.5.0**
---
**Problem:**
EVEREST-1599

**CHECKLIST**
---
**Jira**

**Tests**


controllers/providers/pg/applier.go
```diff
@@ -308,16 +317,7 @@ func defaultSpec(db *everestv1alpha1.DatabaseCluster) pgv2.PerconaPGClusterSpec
...
-			Enabled: false,
-			Resources: corev1.ResourceRequirements{
-				Limits: corev1.ResourceList{
-					corev1.ResourceMemory: resource.MustParse("300M"),
-					corev1.ResourceCPU:    resource.MustParse("500m"),
-				},
-			},
-			Secret: fmt.Sprintf("%s%s-pmm", common.EverestSecretsPrefix, db.GetName()),
-		},
+		PMM: nil,
```
oksana-grishchenko, line 320: Am I understand it correctly that the idea of this refactoring is to unify the `defaultSpec()` signature among the operators? If so, then perhaps it would be more predictable to keep the PMM config inside the default spec as the other operators do, but define only the `Secret` outside. That would make the difference between operators more prominent. 
Just a suggestion, feel free to resolve it if not addressed

oksana-grishchenko (approved):

controllers/providers/pg/applier.go
recharte: Actually that was not the reason. The PGO now enforces that one of the fields in this PMMSpec struct is mandatory and we weren't setting it if the monitoring was turned off. In order to fix it, I decided to remove the struct altogether if monitoring is disabled.

mayankshah1607 (approved):

PR #566, opened 2024-10-29
EVEREST-107 restructure tests and fix DB reconciliation when BackupStorage is changed

recharte:
**Restructure tests and fix DB reconciliation when BackupStorage is changed**
---
**Problem:**
EVEREST-107

Our e2e tests are messy. There are a lot of tests that are not e2e, they don't wait for the DB operators to reconcile.

**Tests changes:**
* Separate the real e2e tests from the integration tests.
* Improve the upgrade tests to actually check that the DB isn't restarted when the operator is upgraded but is correctly restarted and comes up after the CRVersion is updated.
* Remove hardcoded version checks allowing for easier operator upgrades 
* Don't run the DB operators in the integration tests at all. These only assert the CRs get created correctly so no need to have the DB operators running in the background trying to reconcile stuff we don't care about.

**Fixes:**
* Reconcile DB when DBB's BackupStorage is changed
* Reconcile DB when PITR's BackupStorage is changed

**CHECKLIST**
---
**Jira**

**Tests**


oksana-grishchenko (approved):
Thank you for putting e2e tests in order! LGTM.

PR #567, opened 2024-10-31
EVEREST-1598 Update pbm startingStatus interval

oksana-grishchenko:
**Update default PBM startingStatus for new clusters**
---
**Problem:**
EVEREST-1598

The error `couldn''t get response from all shards: convergeClusterWithTimeout: 33s:` can appear for sharded clusters (see EVEREST-1594). There is no certainty why exactly it appears, but the more shards the cluster has the bigger is the chance to face that limit. 
As discussed with experts, it would make sense to increase the default timeout. There's no strict recommendations which new default value to use, so it was decided to use 120s as there were a couple of cases when it helped in the past. 

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/e2e-tests/tests/features/dbbackup_psmdb/10-assert.yaml
+++ b/e2e-tests/tests/features/dbbackup_psmdb/10-assert.yaml
@@ -32,6 +32,12 @@ kind: PerconaServerMongoDB
 metadata:
   name: test-psmdb-cluster
 spec:
+  backup:
+    configuration:
+      backupOptions:
+        oplogSpanMin: 0
+        timeouts:
+          startingStatus: 120
   multiCluster:
     enabled: false
 status:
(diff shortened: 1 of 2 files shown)


e2e-tests/tests/features/dbbackup_psmdb/10-assert.yaml
```diff
@@ -32,6 +32,12 @@ kind: PerconaServerMongoDB
 metadata:
   name: test-psmdb-cluster
 spec:
+  backup:
+    configuration:
+      backupOptions:
+        oplogSpanMin: 0
```
recharte, line 38: Why is this value 0?

oksana-grishchenko: the `.backup.oplogSpanMin` is set to 0 as a side effect, because we need to have the field since we configure the backup options object for PBM. However the 0 value [is ignored by pbm](https://github.com/percona/percona-backup-mongodb/blob/2a6b65bbe29bb4dadadfd9d84ceeea27aced26bd/pbm/config/config.go#L170), so this `oplogSpanMin` does not introduce any changes.

recharte (approved):

PR #581, opened 2024-11-19
EVEREST-107 | monitoringconfig controller improvements and refactor

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-107

Improvements for monitoringconfig controller:
- Use the structs provided by the VictoriaMetrics operator Go package to handle monitoringconfigs. Earlier we were using unstructured objects
- Add a watch for VMAgents so that the controller is able to reconcile it back in a timely manner
- Fix a bug with setting `verifyTLS` in the remote write setting - setting `verifyTLS=false` for one monitoring config results in setting that value for all remote write configs rather than for one particular remote write
- Simplify the clean-up logic - use finalizer to ensure that VMAgent spec is cleaned up when MonitoringConfigs are deleted.
- Add a integration test to verify the VMAgent reconciliation.

**CHECKLIST**
---
**Jira**

**Tests**


go.mod
```diff
@@ -23,12 +24,15 @@ require (
 	gotest.tools v2.2.0+incompatible
 	k8s.io/api v0.31.2
 	k8s.io/apimachinery v0.31.2
-	k8s.io/client-go v0.31.1
+	k8s.io/client-go v12.0.0+incompatible
```
mayankshah1607, line 27: This comes from the victoria-metrics package. To fix this I have pinned the dependencies below in the `replace` block

go.mod
```diff
@@ -119,3 +138,11 @@ require (
...
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace (
+	k8s.io/api => k8s.io/api v0.31.1
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.31.1
+	k8s.io/apimachinery => k8s.io/apimachinery v0.31.1
+	k8s.io/client-go => k8s.io/client-go v0.31.1
+	k8s.io/component-base => k8s.io/component-base v0.31.1
+)
```
michal-kralik, line 148: Can we please move this to the top of the file so it's clear these are replaced?

michal-kralik (approved):

PR #592, opened 2024-11-22
EVEREST-1618: affinity per component

michal-kralik:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1618

Add configuration of affinity per component.
The affinity configuration is forwarded to the respective operator.

**CHECKLIST**
---
**Jira**

**Tests**


api/v1alpha1/databasecluster_types.go
```diff
@@ -349,8 +349,27 @@ type DatabaseClusterSpec struct {
...
 	Sharding *Sharding `json:"sharding,omitempty"`
+	// Affinity defines scheduling affinity per component.
+	Affinity []*DatabaseClusterAffinity `json:"affinity,omitempty"`
 }
 
+type DatabaseClusterAffinity struct {
+	// Component defines the component to which the affinity applies.
+	Component DatabaseClusterAffinityComponent `json:"type"`
+	// Affinity defines scheduling affinity.
+	Affinity corev1.Affinity `json:"affinity,omitempty"`
```
recharte, line 360: This approach also works but I think it would be tidier if instead of having this `DatabaseClusterAffinity` struct with a component field we'd instead take this affinity field and added it to each component's struct (`Engine`, `Proxy`, `Sharding.ConfigServer`). 
Just like you can configure the resources for Engine and Proxy and the Resources field is part of those structs. I envisioned the Affinity to also be part of those structs. This is what I mean:
```
// Engine is the engine configuration.
type Engine struct {
	// Type is the engine type
	// +kubebuilder:validation:Enum:=pxc;postgresql;psmdb
	Type EngineType `json:"type"`
	// Version is the engine version
	Version string `json:"version,omitempty"`
	// Replicas is the number of engine replicas
	// +kubebuilder:validation:Minimum:=1
	Replicas int32 `json:"replicas,omitempty"`
	// Storage is the engine storage configuration
	Storage Storage `json:"storage"`
	// Resources are the resource limits for each engine replica.
	// If not set, resource limits are not imposed
	Resources Resources `json:"resources,omitempty"`
	// Config is the engine configuration
	Config string `json:"config,omitempty"`
	// UserSecretsName is the name of the secret containing the user secrets
	UserSecretsName string `json:"userSecretsName,omitempty"`
	// CRVersion is the desired version of the CR to use with the
	// underlying operator.
	// If unspecified, everest-operator will use the same version as the operator.
	//
	// NOTE: Updating this property post installation may lead to a restart of the cluster.
	// +optional
	CRVersion *string `json:"crVersion,omitempty"`
	// Affinity defines scheduling affinity.
	Affinity corev1.Affinity `json:"affinity,omitempty"`
}

// Proxy is the proxy configuration.
type Proxy struct {
	// Type is the proxy type
	// +kubebuilder:validation:Enum:=mongos;haproxy;proxysql;pgbouncer
	Type ProxyType `json:"type,omitempty"`
	// Replicas is the number of proxy replicas
	// +kubebuilder:validation:Minimum:=1
	// Config is the proxy configuration
	Config string `json:"config,omitempty"`
	// Expose is the proxy expose configuration
	Expose Expose `json:"expose,omitempty"`
	// Resources are the resource limits for each proxy replica.
	// If not set, resource limits are not imposed
	Resources Resources `json:"resources,omitempty"`
	// Affinity defines scheduling affinity.
	Affinity corev1.Affinity `json:"affinity,omitempty"`
}

// ConfigServer represents the sharding configuration server settings.
type ConfigServer struct {
	// Replicas is the amount of configServers
	// +kubebuilder:validation:Minimum:=1
	Replicas int32 `json:"replicas"`
	// Affinity defines scheduling affinity.
	Affinity corev1.Affinity `json:"affinity,omitempty"`
}
```

michal-kralik: I also thought about this approach at first but then figured there's more than 3 affinity configurations per engine.
For example MongoDB has 5.

I was thinking this may be a more flexible and easier to understand approach for users to managing affinity?
I'm happy to put it into the 3 structs though if that's preferred.

recharte: I see your point, however with the current API we can't even enable the nonvoting/arbiter nodes so we don't really need to consider setting affinity for them. If we ever add support for configuring the nonvoting/arbiter nodes we'll need to expose it somehow which can then also have the affinity configuration there.

> easier to understand approach for users

I actually disagree here, I think that it's easier to understand an `Affinity` field embedded in `Engine`, `Proxy`, `Sharding.ConfigServer` than to see an `Affinity` array at the root of the spec where you then need to know how to name each component in order to fill in `Affinity.Component`.
Let's hear from @oksana-grishchenko and @mayankshah1607 maybe they feel differently than me :slightly_smiling_face: I'm open to changing my mind.

mayankshah1607: I agree with Diogo, placing it into its 3 respective structs is also consistent with the current spec

michal-kralik: updated

controllers/providers/pxc/applier.go
```diff
@@ -126,6 +126,9 @@ func (p *applier) Engine() error {
 	if p.DB.Status.Status == everestv1alpha1.AppStateReady {
 		pxc.Spec.PXC.PodSpec.Affinity = p.currentPerconaXtraDBClusterSpec.PXC.Affinity
 	}
+
+	p.configureEngineAffinity()
```
recharte, line 130: I think we need to do some cleanup above because we were setting some default affinity that we got from `common.DefaultAffinitySettings`.
Besides cleaning up those defaults, we should also ensure that we don't change the affinity for existing cluster (see the if condition 4 lines above this comment) otherwise all DB clusters would restart when upgrading the everest operator to 1.4.

recharte (commented):
@michal-kralik since this PR will apply the affinity configuration that we get from the UI we need to hold this PR until the FE has also implemented this change :slightly_smiling_face:

controllers/providers/pxc/applier.go
michal-kralik: I'm not sure I follow.
Re restarts - the `p.configureEngineAffinity()` triggers a change only in case you configure affinity configuration for an `Engine`.
Since the spec does not support affinity at the moment, existing clusters shall not be affected.

Re cleanup - the line https://github.com/percona/everest-operator/blob/3a0cc19738d8fc4611a89e3be54c98433640e95c/controllers/providers/pxc/affinity.go#L24 cleans up previous affinity settings. Is there anything else you had in mind?

michal-kralik:
> since this PR will apply the affinity configuration that we get from the UI we need to hold this PR until the FE has also implemented this change 🙂 (Disabled auto-merge)

The change will not have any effect if an affinity configuration is not provided. Without the UI change, the behavior will be the same regardless if the PR gets merged.
Why do you think it's important we wait for the UI?

controllers/providers/pxc/applier.go
michal-kralik: You're right.
I thought it's better not to restart the DB and keep the behavior consistent across minor versions.
Happy to change this in a way which makes more sense. It doesn't seem to be possible to determine if the default affinity settings are expected if affinity is not provided.

api/v1alpha1/databasecluster_types.go
```diff
@@ -326,6 +330,8 @@ type Sharding struct {
 	Shards int32 `json:"shards"`
 	// ConfigServer represents the sharding configuration server settings
 	ConfigServer ConfigServer `json:"configServer"`
+	// Affinity defines scheduling affinity.
+	Affinity *corev1.Affinity `json:"affinity,omitempty"`
```
recharte, line 334: This should be moved inside the ConfigServer struct as it only applies to that component instead of the whole sharding config.

controllers/providers/psmdb/affinity.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, rs := range p.PerconaServerMongoDB.Spec.Replsets {
+		if rs.Affinity == nil {
+			rs.Affinity = &psmdbv1.PodAffinity{}
+		}
+		rs.Affinity.Advanced = affinity
+	}
+}
+
+func (p *applier) configureShardingAffinity() {
+	affinity := p.DB.Spec.Engine.Affinity
```
recharte, line 35:
```suggestion
	affinity := p.DB.Spec.Sharding.ConfigServer.Affinity
```

michal-kralik: updated

api/v1alpha1/databasecluster_types.go
michal-kralik: updated

tests/e2e/core/pg/10-create-cluster.yaml
```diff
@@ -14,6 +14,26 @@ spec:
...
+    affinity:
+      nodeAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 5
+          preference: 
+            matchExpressions:
+            - key: another-node-label-key
+              operator: In
+              values:
+              - another-node-label-value
```
recharte, line 26: I'm commenting here but this is a generic comment for all tests added in this PR.
Even though we are setting this node affinity as preferred which shouldn't impact the scheduling of the pods, I'd still move these tests to the `tests/integration/features` directory because we don't need to complicate the core e2e tests with this. Since we only want to assert that the everest operator correctly translated the affinity settings to the pxc/psmdb/pg CRs, we don't need the pxc/psmdb/pg operators to be running, IMO the affinity tests should be run as an integration test and we'd avoid polluting the core e2e tests with some affinity settings that can interfere with the results.

tests/integration/features/affinity/02-psmdb.yaml
```diff
@@ -0,0 +1,47 @@
...
+    affinity:
+      nodeAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 5
+          preference: 
+            matchExpressions:
+            - key: another-node-label-key
+              operator: In
+              values:
+              - another-node-label-value
```
recharte, line 47: Can we change this to a sharded cluster and set the cfgsrv affinity so that we also cover that case?

michal-kralik: updated

recharte (approved):

PR #603, opened 2024-12-12
EVEREST-1689 | Use new PSMDB finalizers

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1689

In PSMDB operator 1.17.0, the finalizer names were changed to to contain fully qualified domain names to comply with kubernetes standards. This PR updates the finalizers to use the new names

**CHECKLIST**
---
**Helm chart**
~- [ ] Is the [helm chart](https://github.com/percona/percona-helm-charts/tree/main/charts/everest) updated with the new changes? (if applicable)~

**Jira**

**Tests**


internal/controller/databaseclusterbackup_controller.go
```diff
@@ -74,7 +74,7 @@ const (
 	pxcGapsReasonString               = "BinlogGapDetected"
 
 	deletePXCBackupFinalizer   = "delete-s3-backup"
-	deletePSMDBBackupFinalizer = "delete-backup"
+	deletePSMDBBackupFinalizer = "percona.com/delete-backup"
```
michal-kralik, line 77: Could this change cause any issues during upgrades?
i.e. if an existing cluster does not have the correct finalizer yet?

mayankshah1607: > i.e. if an existing cluster does not have the correct finalizer yet?

I think that's fine, PSMDB operator will recognize the old finalizer as well, until 1.20.0. So both types of finalizers should work...

ref: https://github.com/percona/percona-server-mongodb-operator/releases/tag/v1.17.0

michal-kralik: would we need to take any action with 1.20.0? If yes, shall there be a ticket for that?

mayankshah1607: I don't think so.. Once users upgrade to Everest 1.5.0, they will move to the new finalizers. So no action is required by the time we support PSMDBO 1.20. Am I missing something?

michal-kralik: I'm not sure. Wouldn't the old DB be stuck with the old finalizer which is not managed by anyone?
Maybe a test would be good for this.

Another topic - could this change trigger an unwanted restart of a DB?

mayankshah1607: > Wouldn't the old DB be stuck with the old finalizer which is not managed by anyone?

Right, yeah. I assumed that the PSMDB operator would remove the old one, but that's not the case. I just updated the code to remove the old finalizers and replace them with the new ones.

> Maybe a test would be good for this.

I'm not sure how you would write a test for this. To have a reliable test, you would need to provision a DB with an older (v1.4.0) everest-operator, then upgrade the everest-operator to the current one, and then check that old finalizers don't exist.. our tests run only 1 version of the operator, and that's the current one. Do you have any ideas to achieve this easily?

> Another topic - could this change trigger an unwanted restart of a DB?

It won't, since the finalizers are applied on the PDMSB CR, and do not affect the StatefulSet pods. (tested it anyway to be extra sure)

michal-kralik: re the tests - in the worst case we could test at least manually if it works ok.
But maybe we could create the DB, add finalizers directly in the PSMDB CR and trigger a reconciliation to check if they're removed properly? Not sure, haven't tried this approach yet.

oksana-grishchenko (approved):

PR #605, opened 2024-12-17
EVEREST-107 run e2e tests conditionally

recharte:
The E2E test suite takes a long time to run and these tests are not always needed.

This PR introduces a condition for running the E2E test suite based on PR labels. There are 3 labels relevant:
* `test/e2e` - runs the full e2e test suite (core+upgrade)
* `test/e2e-core` - runs just the core e2e test suite
* `test/e2e-upgrade` - runs just the upgrade e2e test suite

Besides the conditional run of E2E tests, this PR introduces some more changes to the integration tests:
* All integration tests are now run in kind. Since they don't wait for any status, there's no point in running them on EKS. This will save us time and money.
* Adds new core tests to cover the basic scenarios
* Moves the dbengine tests to a separate step to avoid having them interfere with the rest of the tests.


.github/workflows/ci.yml
```diff
@@ -46,9 +47,82 @@ jobs:
           PG_OPERATOR_VERSION: 2.5.0
           PERCONA_VERSION_SERVICE_URL: https://check-dev.percona.com/versions/v1
 
+  integration-tests:
```
mayankshah1607, line 50: I was expecting a step to run `kind create cluster ...`, but that's not there. When does the kind cluster get created? 🤔

mayankshah1607 (approved):

.github/workflows/ci.yml
```diff
@@ -46,9 +47,82 @@ jobs:
...
+          PSMDB_OPERATOR_VERSION: 1.18.0
+          PG_OPERATOR_VERSION: 2.5.0
+          PREVIOUS_PG_OPERATOR_VERSION: 2.4.1
+          PREVIOUS_PXC_OPERATOR_VERSION: 1.14.0
+          PREVIOUS_PSMDB_OPERATOR_VERSION: 1.17.0
+          PERCONA_VERSION_SERVICE_URL: https://check-dev.percona.com/versions/v1
 
   eks-e2e-tests:
     name: E2E tests (EKS)
+    if: contains(github.event.pull_request.labels.*.name, 'test/e2e') || contains(github.event.pull_request.labels.*.name, 'test/e2e-upgrade') || contains(github.event.pull_request.labels.*.name, 'test/e2e-core')
```
oksana-grishchenko, line 125: The E2E step is a required check in CI, so we skip it, the PR will be blocked, wouldn't it? Should we configure Merge Gatekeeper then?

oksana-grishchenko: Although now it looks like the PR is not blocked by the skipped step. Not sure😅

.github/workflows/ci.yml
recharte: This is handled internally by kuttl if you set the `StartKIND` field.
https://github.com/percona/everest-operator/blob/765cb75a70909d21cf1c5175f2186abb86a55acb/tests/integration/kuttl-features.yaml#L5

.github/workflows/ci.yml
recharte: I was concerned about this as well but I didn't check how the merge gatekeeper works internally because I also saw that the PR wasn't blocked by it. It seems gatekeeper is smart enough to know that a skipped job should not be taken into account.

oksana-grishchenko (approved):

PR #615, opened 2024-12-20
EVEREST-1549 PXC 1.16.1

oksana-grishchenko:
**Add PXC 1.16.0 support**
---
**Problem:**
EVEREST-1549

Add PXC 1.16.0 support

**Related pull requests**
- https://github.com/percona/everest/pull/953
- https://github.com/percona/everest-catalog/pull/52

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**


internal/controller/providers/pxc/applier.go
```diff
@@ -49,7 +49,13 @@ func (p *applier) Paused(paused bool) {
...
 
 func (p *applier) AllowUnsafeConfig(allow bool) {
-	p.PerconaXtraDBCluster.Spec.AllowUnsafeConfig = allow
+	p.PerconaXtraDBCluster.Spec.AllowUnsafeConfig = false
+	p.PerconaXtraDBCluster.Spec.Unsafe = pxcv1.UnsafeFlags{
+		TLS:               false,
+		PXCSize:           allow,
+		ProxySize:         allow,
+		BackupIfUnhealthy: false,
+	}
```
oksana-grishchenko, line 58: Changes in PXC 1.16 causes a change in the PXC StatefulSet if the old `AllowUnsafeConfig` is used, which means a restart for 1-node clusters. To avoid the restart we need to set the `AllowUnsafeConfig` always explicitly to false and use more specific `UnsafeFlags` instead.

tests/e2e/core/pxc/60-assert.yaml
```diff
@@ -65,6 +59,9 @@ spec:
           requests:
             storage: 15G
   secretsName: pxc-sample-secrets
+  unsafeFlags:
+    proxySize: true
+    pxcSize: true
```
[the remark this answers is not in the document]
recharte: I know that this was logic wasn't introduced by you in this PR but does it still make sense now that we have separate unsafe flags? I believe that we only had this logic because the allowUnsafeConfig flag also disabled TLS and because of that we couldn't scale properly scale up the cluster if we disabled allowUnsafeConfig... I don't remember the details but since we are only setting the size flags I don't that's is needed anymore...

internal/controller/providers/pxc/applier.go
recharte: When we upgade the everest operator all 1-node PXC clusters will see the TLS and BackupIfUnhealthy flags change from true to false. Will this lead to any issues or restarts? IIRC issues with disabling these flags was the reason why we couldn't set the AllowUnsafeConfig to false when scaling up clusters...

recharte (approved):

internal/controller/providers/pxc/applier.go
oksana-grishchenko: The `TLS` and `BackupIfUnhealthy` flags are actually never set to true in Everest 1.4.0 with PXC 1.15.0. The pxc operator [tries to set them](https://github.com/percona/percona-xtradb-cluster-operator/blob/main/pkg/apis/pxc/v1/pxc_types.go#L914-L923) under the hood when `AllowUnsafeConfig` flag is used, however I don't see those flags in the actual pxc CRs of 1 node:
Screenshot 2025-01-20 at 16 38 37

 I think it happens because Everest [applies the default spec](https://github.com/percona/everest-operator/blob/f027a586265c45663b81b697c5ca390645c674cc/internal/controller/providers/pxc/provider.go#L100) and then reconciles what it knows, and it never knew about that `UnsafeFlags` section so the section is gone. 

I checked the sts changes when upgrading from Everest 1.4.0 to this new version which has
```
		TLS:               false,
		PXCSize:           allow,
		ProxySize:         allow,
		BackupIfUnhealthy: false,
```
the sts wasn't changed, the restart did not happen. 


> IIRC issues with disabling these flags was the reason why we couldn't set the AllowUnsafeConfig to false when scaling up clusters

I can't remember this discussion about scaling up, however since the flags were never set on the PXC side, setting AllowUnsafeConfig to false shouldn't be a problem... but I may have missed something here. Maybe it was a problem when upgrading existing clusters from pxc 1.14 to 1.15🤔 I'm a bit confused, it would be great if you would share some more details about that scaling up problem.

tests/e2e/core/pxc/60-assert.yaml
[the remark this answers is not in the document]
oksana-grishchenko: > Should these still be set to true? This is now a 3-node cluster...
Technically when we scale the cluster up it shouldn't have the unsafe flags set to true anymore, however we never "unset" the AllowUnsafeConfig when we were scaling up, so I kept this behaviour. 

I agree it makes sense to change it. We need to understand when we try to scale up and simultaneously, in the same reconciliation loop, set the`AllowUnsafeConfig`, `UnsafeFlags.PXCSize` and `UnsafeFlags.ProxySize` to false. That should work.

internal/controller/providers/pxc/applier.go
recharte: Sorry, I mislead you :see_no_evil: 
> all 1-node PXC clusters will see the TLS and BackupIfUnhealthy flags change from true to false

This is not what I meant to say :see_no_evil: I know that the flags weren't set. I actually meant the internal state of them in the PXCO.
In v1.4.0, we were setting `AllowUnsafeConfig: true` which triggers the PXCO to internally disable TLS.
By upgrading to v1.5.0 we are now setting `AllowUnsafeConfig: false` and `UnsafeFlags.TLS: false`, I assume that this will trigger PXCO to enable TLS on existing clusters.

The scaling fix was introduced [here](https://github.com/percona/everest-operator/pull/57).
There's [this comment](https://github.com/percona/everest-operator/pull/57#issuecomment-1702787810) from Andrew:
> Immutability for unsafe options is made for now because not all upstream operators support it. AFAIK it worked perfectly for PXC a while ago but right now it's either broken or I'm mistaken here. PSMDB does not support it and it should be immutable and I don't know anything about PG right now.


I don’t fully remember the exact issue with scaling up, but I believe it was related to the 1-node cluster having TLS disabled (set internally by `AllowUnsafeConfig: true`). When we scaled up to 3 nodes, we also changed `AllowUnsafeConfig` to `false`, which caused PXCO to fail because it couldn’t handle the TLS configuration change properly. To avoid it, we decided to keep `AllowUnsafeConfig: true` when scaling up.

oksana-grishchenko: After a discussion it was decided to deprecate the `allowUnsafeConfiguration` flag in the everest-operator API and only support it for old clusters where it is already defined (for backward compatibility), the PR is updated.  

Test results for the current implementation:
Screenshot 2025-01-22 at 11 05 32

tests/e2e/core/pxc/60-assert.yaml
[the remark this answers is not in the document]
oksana-grishchenko: The implementation is updated, the update scenario for scaled up clusters works without restarts, see the https://github.com/percona/everest-operator/pull/615#discussion_r1924974826 for configuration details.

recharte (approved):

recharte (approved):

PR #650, opened 2025-02-04
EVEREST-1841 MongoDB Major Version Upgrades

recharte:
**MongoDB Major Version Upgrades**
---
**Problem:**
EVEREST-1841

In order to get a fully automated major version upgrade of MongDB we need to set FCV. This automatically enables the features of the new version after upgrading. The drawback is that we can't downgrade but we are blocking downgrades regardless. More info about the MongoDB major version upgrades can be found [here](https://docs.percona.com/percona-operator-for-mongodb/update.html#major-version-automated-upgrades). 

I also extended the E2E tests to have a DB upgrade suite (triggered by the `test/e2e-db-upgrade` label). By doing so I changed the label that triggers the operator upgrade tests from `test/e2e-upgrade` to `test/e2e-operator-upgrade`.

**Related pull requests**

- https://github.com/percona/everest/pull/1080

**CHECKLIST**
---
**Helm chart**
- ~[ ] Is the [helm chart](https://github.com/percona/percona-helm-charts/tree/main/charts/everest) updated with the new changes? (if applicable)~

**Jira**

**Tests**


maxkondr (approved):

.github/workflows/ci.yml
```diff
@@ -280,6 +280,52 @@ jobs:
           kubectl delete apiservice v1.packages.operators.coreos.com --ignore-not-found=true
           kubectl get crd -o name | grep .coreos.com$ | awk -F '/' {'print $2'} | xargs --no-run-if-empty kubectl delete crd
 
+      - name: Run db upgrade tests
+        if: contains(github.event.pull_request.labels.*.name, 'test/e2e') || contains(github.event.pull_request.labels.*.name, 'test/e2e-db-upgrade')
+        run: |
+          kubectl kuttl test --config ./tests/e2e/kuttl-db-upgrade.yaml
+        env:
+          PXC_OPERATOR_VERSION: 1.16.1
+          PSMDB_OPERATOR_VERSION: 1.19.0
```
oksana-grishchenko, line 289: should we use `1.19.1` since it's the latest supported version?

oksana-grishchenko (approved):

maxkondr (approved):

.github/workflows/ci.yml
```diff
@@ -280,6 +280,52 @@ jobs:
           kubectl delete apiservice v1.packages.operators.coreos.com --ignore-not-found=true
           kubectl get crd -o name | grep .coreos.com$ | awk -F '/' {'print $2'} | xargs --no-run-if-empty kubectl delete crd
 
+      - name: Run db upgrade tests
+        if: contains(github.event.pull_request.labels.*.name, 'test/e2e') || contains(github.event.pull_request.labels.*.name, 'test/e2e-db-upgrade')
+        run: |
+          kubectl kuttl test --config ./tests/e2e/kuttl-db-upgrade.yaml
+        env:
+          PXC_OPERATOR_VERSION: 1.16.1
```
tplavcic, line 288: Is this something which needs to be updated after each release?

tplavcic (changes requested):

.github/workflows/ci.yml
recharte: Yeah, [we already use these env vars for the other e2e tests](https://github.com/percona/everest-operator/blob/a36bc725e5792cd1e7bbeae49ba682a514702ae3/.github/workflows/ci.yml#L88-L119).
As you identified, it's not great because we need to update them every time we add a new operator version, e.g. [PGO v2.6.0](https://github.com/percona/everest-operator/pull/696/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f).
However, this is already a problem, not something that we are introducing with this PR so maybe we shouldn't block it . Instead, we should open a ticket to address this.

tplavcic: One central place to store these versions could be ok (if not possible to determine dynamically), but the bigger problem here is that all assertion files and test files have fixed DB and image versions inside, so once you need to change the version you need to do it in many places.
https://github.com/percona/everest-operator/pull/650/files#diff-e1404df47dae19d052286e694b5819ce5c2856027879328b82de2616055e1809R14
https://github.com/percona/everest-operator/pull/650/files#diff-6b9fc8d2eba2b9fbd1fb615972eeb2f196b96cc38d070a575c0cafae5a9c9e3bR41

Also consider maybe to not store it in the CI workflow file, but instead in something inside the test suite like: https://github.com/percona/percona-server-mysql-operator/blob/main/e2e-tests/vars.sh

Don't get me wrong I don't think we should block this if there is no time to fix it, I will open a ticket for improvement.

recharte: Yeah you're right, the hardcoded versions in the test files is a lot more problematic. I'm not sure how to circumvent that thought, I don't see an easy way to dynamically set these versions. Let me try to think a little bit deeper about this and check what can be done.

tplavcic: It can be done like this: https://github.com/percona/percona-server-mysql-operator/blob/main/e2e-tests/tests/gr-demand-backup/02-create-cluster.yaml
Here only namespace is dynamic, but you get the idea.

As for assertions some part can be put as fixed with skipped dynamic fields and the dynamic fields can be checked with a script inside assertion like: https://github.com/percona/percona-server-mysql-operator/blob/main/e2e-tests/tests/gr-demand-backup/01-assert.yaml#L26

tplavcic: Opened: https://perconadev.atlassian.net/browse/EVEREST-1938

tplavcic (approved):

.github/workflows/ci.yml
recharte: Thank you @tplavcic. We'll probably need to write a script to query the version service and build an appropriate set of versions that can be used here. Currently, we need 2 minor versions and 2 major version upgrades.

PR #653, opened 2025-02-05
EVEREST-1814 add creating state

oksana-grishchenko:
**Add "Creating" DB status**
---
**Problem:**
EVEREST-1814

**Problem:**
The DB can be in an ‘unknown’ state for a couple of seconds and we didn't want to show that to the user. 

**Solution:**

The upstream operators have the different statuses when a cluster is being created: 
- pxc - "unknown"
- psmdb - ""
- pg does not have any empty status.

So the upstream statuses are different although Everest maps the DB status to the upstream cluster status 1:1. There is no point so far to create a separate mapping for each upstream operator only because we want to unify AppStateCreating, so in this PR the Everest operator unifies the "" and "unknown" statuses into the "creating" status. 

**Related PRs:** 
- https://github.com/percona/everest/pull/1084

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/api/v1alpha1/databasecluster_types.go
+++ b/api/v1alpha1/databasecluster_types.go
@@ -33,6 +33,8 @@ var (
 const (
 	// AppStateUnknown is an unknown state.
 	AppStateUnknown AppState = "unknown"
+	// AppStateCreating is used when the db did not start to initialize yet.
+	AppStateCreating AppState = "creating"
 	// AppStateInit is a initializing state.
 	AppStateInit AppState = "initializing"
 	// AppStatePaused is a paused state.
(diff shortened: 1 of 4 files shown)


api/v1alpha1/databasecluster_types.go
```diff
@@ -33,6 +33,8 @@ var (
 const (
 	// AppStateUnknown is an unknown state.
 	AppStateUnknown AppState = "unknown"
```
mayankshah1607, line 35: should we remove this since it is not used anywhere now?

maxkondr: in addition - is AppStateNew used somehow? and shall it be replaced with AppStateCreating as well?

oksana-grishchenko: The way we map statuses now is `everestv1alpha1.AppState(upstream.Status.State)` which means we still need both 

AppStateUnknown AppState = "unknown"
AppStateNew AppState = ""

because pxc has "unknown", psmdb has "", pg does not have any empty status.

As an alternative we could remove the statuses^ and have a `switch` mapping in each provider separately. I thought it's not really needed right now because the other statuses are mapping 1:1. Do you think it would be better to have the mapping for each upstream operator? I agree it's a bit confusing to have the statuses that are used only internally, but not sure which option would be better here.

maxkondr (approved):

api/v1alpha1/databasecluster_types.go
mayankshah1607: > have a switch mapping in each provider separately

I can't remember clearly, but I think we used to have this earlier and we removed it for some reason.

But anyway, I'm fine with keeping them around since they are used in upstream operators. My concern was that these vars are not used anywhere currently, but its probably fine to keep them

mayankshah1607 (approved):

PR #662, opened 2025-02-11
EVEREST-1849 | Do not set owner reference on PSMDB user secret

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1849

Fixes the issue with PSMDB clusters restarting upon deletion.

**Cause:**
- we set an owner reference on the user secret created by the everest-operator
- upon deletion of the cluster, this user secret is automatically deleted first due to the owner reference
- once the secret is deleted, the psmdb-operator recreates with new default information
- since the secret was technically changed, the psmdb-operator needs to restart the statefulset pods in order to propagate the new information, leading to restarts

**Solution:**
Do not set the owner reference on the secret. The psmdb-operator shall anyway handle the secret clean-up as a part of the `delete-psmdb-pvc` finalizer. This way, the secret is deleted by the psmdb-operator only after the pods are deleted, hence avoiding any unnecessary restarts.

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/internal/controller/providers/psmdb/applier.go
+++ b/internal/controller/providers/psmdb/applier.go
@@ -370,6 +370,7 @@ func (p *applier) applyPMMCfg(monitoring *everestv1alpha1.MonitoringConfig) erro
 		map[string][]byte{
 			"PMM_SERVER_API_KEY": []byte(apiKey),
 		},
+		false,
 	); err != nil {
 		return err
 	}
--- a/internal/controller/common/helper.go
+++ b/internal/controller/common/helper.go
@@ -262,6 +262,7 @@ func CreateOrUpdateSecretData(
 	database *everestv1alpha1.DatabaseCluster,
 	secretName string,
 	data map[string][]byte,
+	setControllerRef bool,
 ) error {
 	err := UpdateSecretData(ctx, c, database, secretName, data)
 	if err == nil {
@@ -280,10 +281,14 @@ func CreateOrUpdateSecretData(
 		Type: corev1.SecretTypeOpaque,
 		Data: data,
 	}
-	err = controllerutil.SetControllerReference(database, secret, c.Scheme())
-	if err != nil {
-		return err
+
+	if setControllerRef {
+		err = controllerutil.SetControllerReference(database, secret, c.Scheme())
+		if err != nil {
+			return err
+		}
 	}
+
 	// If the secret does not exist, create it
 	return c.Create(ctx, secret)
 }
--- a/internal/controller/providers/pg/applier.go
+++ b/internal/controller/providers/pg/applier.go
@@ -309,10 +309,16 @@ func (p *applier) applyPMMCfg(monitoring *everestv1alpha1.MonitoringConfig) erro
 		return err
 	}
 
+	// We avoid setting the controller reference for the Secret here.
+	// The PSMDB operator handles Secret cleanup automatically as part of the `delete-psmdb-pvc` finalizer process.
+	// If we were to set the controller reference, the Secret would be deleted before the PSMDB cluster, leading to
+	// the PSMDB operator recreating the Secret and restarting the DB pods to sync updated user information,
+	// which would cause unnecessary restarts.
 	if err := common.CreateOrUpdateSecretData(ctx, c, database, pg.Spec.PMM.Secret,
 		map[string][]byte{
 			"PMM_SERVER_KEY": []byte(apiKey),
 		},
+		true,
 	); err != nil {
 		return err
 	}


internal/controller/providers/psmdb/applier.go
```diff
@@ -366,10 +366,17 @@ func (p *applier) applyPMMCfg(monitoring *everestv1alpha1.MonitoringConfig) erro
 		return err
 	}
 
+	// We avoid setting the controller reference for the Secret here.
+	// The PSMDB operator handles Secret cleanup automatically as part of the `delete-psmdb-pvc` finalizer process.
+	// If we were to set the controller reference, the Secret would be deleted before the PSMDB cluster, leading to
+	// the PSMDB operator recreating the Secret and restarting the DB pods to sync updated user information,
+	// which would cause unnecessary restarts.
+	setControllerRef := false
```
mayankshah1607, line 374: This variable is not actually required, we can simply pass `false` to the function parameter, but I added it for the sake of readability only - there are many parameters in the function call and its not immediately obvious that the `false` refers to the controller reference setting..

recharte (commented):
The changes look good but is there a way for us to detect these restarts in tests? Maybe in the e2e/core...

recharte:
@mayankshah1607 I think that there might be more to the restarts than just the secret's owner reference, when running the e2e/core tests I still see 5 or 6 new generations of the PSMDB CR after setting the deletion timestamp, there's some funky reconciliation going on...

mayankshah1607:
> when running the e2e/core tests I still see 5 or 6 new generations of the PSMDB CR after setting the deletion timestamp, there's some funky reconciliation going on...

Yes, I'm aware of that, but its not coming from the everest-operator. It comes from the PSMDB operator itself, so there's nothing we can do about it. Moreover, those spec changes aren't leading to pod restarts, so I think it should be fine.

@tplavcic ran this PR in his E2E tests already, so he can confirm this

recharte:
> > when running the e2e/core tests I still see 5 or 6 new generations of the PSMDB CR after setting the deletion timestamp, there's some funky reconciliation going on...
> 
> @recharte Yes, I'm aware of that, but its not coming from the everest-operator. It comes from the PSMDB operator itself, so there's nothing we can do about it (I tested this without Everest as well). Everest operator does not change the PSMDB CR after there's a deletion timestamp on the DB ([see early return](https://github.com/percona/everest-operator/blob/main/internal/controller/databasecluster_controller.go#L138)).. Moreover, those spec changes aren't leading to pod restarts, so I think it should be fine.
> 
> @tplavcic ran this PR in his E2E tests already, so he can confirm this

Okay, that's fine then. Nonetheless, is there something we can change in the tests to make sure that they fail if there is some restart during deletion? i.e. to detect the bug that this PR fixes?

mayankshah1607:
> @mayankshah1607 I think that there might be more to the restarts than just the secret's owner reference, when running the e2e/core tests I still see 5 or 6 new generations of the PSMDB CR after setting the deletion timestamp, there's some funky reconciliation going on...

hmm, I initially thought something like [this](https://github.com/percona/everest-operator/blob/main/tests/e2e/upgrade/pxc/20-upgrade-operator.yaml#L13) would work, but the issue is that the deletion kicks off seconds after the psmdb-operator performs the pod restart. So the "initializing" state is never captured... I don't see a simple way, definitely requires some complex bash/kubectl script 😅

tplavcic:
> > when running the e2e/core tests I still see 5 or 6 new generations of the PSMDB CR after setting the deletion timestamp, there's some funky reconciliation going on...
> 
> @recharte Yes, I'm aware of that, but its not coming from the everest-operator. It comes from the PSMDB operator itself, so there's nothing we can do about it (I tested this without Everest as well). Everest operator does not change the PSMDB CR after there's a deletion timestamp on the DB ([see early return](https://github.com/percona/everest-operator/blob/main/internal/controller/databasecluster_controller.go#L138)).. Moreover, those spec changes aren't leading to pod restarts, so I think it should be fine.
> 
> @tplavcic ran this PR in his E2E tests already, so he can confirm this

I can confirm that in the case of PSMDB sharding test this fix has resolved the issues with the test. I had 5 test runs without this fix and only 1 run was fine and 4 were bad and with this PR all 5 runs were just fine.
In the Jira ticket itself I have mentioned that there are some other fishy object changes, but I have not investigated this right now because it might be too much for one PR.
If there are no objective problems with this PR I would propose to merge it so that we can investigate (possible) other remaining issues since I don't expect that we will be able to fix all at once.

oksana-grishchenko (approved):
> The psmdb-operator shall anyway handle the secret clean-up as a part of the `delete-psmdb-pvc` finalizer. 

So for pxc the secret should be deleted by the `delete-pxc-pvc` finalizer, for pg - `percona.com/delete-pvc` 

Looks good 👍
I'm only wondering why the problem is prominent only for `psmdb`, do you have any idea?

tplavcic:
@recharte @mayankshah1607 If we have a loop somewhere where we wait that the cluster becomes deleted we can maybe check the object generation and if it increases then fail the test. In the playwright we wait for cluster to be removed from the list, but I'm not sure how we could add it there, maybe it would be good to add it to `kuttl` test somewhere.
Our current test did find this issue, but only because deletion took way too long so it's not bulletproof.

mayankshah1607:
> So for pxc the secret should be deleted by the delete-pxc-pvc finalizer, for pg - percona.com/delete-pvc

@oksana-grishchenko I'm actually not sure, I haven't checked. But it seems like a fair assumption...

> why the problem is prominent only for psmdb

actually, the issue occurs only in sharded clusters. My guess is that it is because they take longer to delete..

mayankshah1607:
> check the object generation and if it increases then fail the test

@tplavcic I don't think that's necessarily correct. I see that the generation increases even though there are no restarts. Basically, a change in generation does not necessarily imply a restart..

recharte (approved):

PR #681, opened 2025-03-03
EVEREST-1651 Fix setting BackupStorageName into DatabaseCluster.labels

maxkondr:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1651

For some reason during creating DB from backup Everest operator wrote backup name into DatabaseCluster.labels that lead to too long label key because backup name is not limited.

**Solution:**
Fixed to store in  DatabaseCluster.labels backup storage name instead.


internal/controller/databasecluster_controller.go
```diff
@@ -369,8 +369,24 @@ func (r *DatabaseClusterReconciler) reconcileLabels(
...
-		updated[fmt.Sprintf(backupStorageNameLabelTmpl, database.Spec.DataSource.DBClusterBackupName)] = backupStorageLabelValue
+		// need to obtain backupStorageName by .spec.dataSource.dbClusterBackupName.dbClusterBackupName
+		dbBackup := &everestv1alpha1.DatabaseClusterBackup{}
+		err := r.Get(ctx, types.NamespacedName{
+			Name:      database.Spec.DataSource.DBClusterBackupName,
+			Namespace: database.Namespace,
+		}, dbBackup)
+		if err != nil {
+			return errors.Join(err, fmt.Errorf("could not get DB backup '%s' to obtain backup storage name", database.Spec.DataSource.DBClusterBackupName))
+		}
```
oksana-grishchenko, line 380: the `database.Spec.DataSource.DBClusterBackupName` can be empty (see [here](https://github.com/percona/everest-operator/blob/a3ec50251812e907a7451573fc8c81c72d1ea81a/internal/controller/common/helper.go#L316-L319)), we need to handle it so no error would be returned in that case.

maxkondr: done

oksana-grishchenko: Now the `database.Spec.DataSource.BackupSource.BackupStorageName` (lines 385-387) is handled only if `database.Spec.DataSource.DBClusterBackupName != ""` however the values are independent. 

I know the relations are pretty confusing, let me describe it in more details: 
The `database.Spec.DataSource` is used when a new DB wants to populate itself with some data during creation. So the field could be empty if we want an empty DB. However if the field is not empty it could contain either:  
- `DBClusterBackupName` - the name of the backup CR with the data we want 
- `BackupSource` - the information about the storage (name, path) that contains the actual backup files.

oksana-grishchenko (approved):

PR #688, opened 2025-03-06
EVEREST-1874 | support volume expansion for PXC

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1874

Allow PVC resizing (storage expansion) for PXC clusters:

- set `pxc.spec.enableVolumeExpansion=true` when storage size is increased.
- storage scaling may be prevented by setting `db.spec.engine.storage.disableVolumeExpansion=true`
- add new DB state `resizingVolumes`
- add a new conditions list in status to indicate failures related to storage resizing

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**


internal/controller/common/helper.go
```diff
@@ -793,3 +794,87 @@ func toUnstructured(obj runtime.Object) (*unstructured.Unstructured, error) {
...
+	hasStorageExpanded := params.CurrentSize.Cmp(params.DesiredSize) < 0 && !params.CurrentSize.IsZero()
+	if !hasStorageExpanded {
+		params.SetStorageSizeFunc(params.DesiredSize)
+		return nil
+	}
+
+	allowedByStorageClass, err := storageClassSupportsVolumeExpansion(ctx, c, params.StorageClass)
+	if err != nil {
+		return fmt.Errorf("failed to check if storage class supports volume expansion: %w", err)
+	}
```
oksana-grishchenko, line 852: In case `params.DisableVolumeExpansion` is `true` the information about `allowedByStorageClass` is never used, so we can put this block closer to the `if !allowedByStorageClass` to avoid redundant listing of storage classes

internal/controller/common/helper.go
```diff
@@ -793,3 +794,87 @@ func toUnstructured(obj runtime.Object) (*unstructured.Unstructured, error) {
...
+	DisableVolumeExpansion bool
+	CurrentSize            resource.Quantity
+	SetStorageSizeFunc     func(resource.Quantity)
+}
+
+// ConfigureStorage handles storage configuration and volume expansion checks for the given database cluster.
+func ConfigureStorage(ctx context.Context, c client.Client, db *everestv1alpha1.DatabaseCluster, params ConfigureStorageParams) error {
+	meta.RemoveStatusCondition(&db.Status.Conditions, everestv1alpha1.ConditionTypeCannotExpandStorage)
+
+	hasStorageExpanded := params.CurrentSize.Cmp(params.DesiredSize) < 0 && !params.CurrentSize.IsZero()
```
oksana-grishchenko, line 843: would it make sense to add a condition for the case the user tries to downscale the volume? perhaps ignoring is just fine to not complicate things. just a thought, wanted to bring it to consideration

oksana-grishchenko (approved):
Using conditions to describe the reasons looks good to me. 
The idea of having the expansion enabled by default is also fine, I don't see any issues so far we could possibly face with it. 
Approving the PR, left a couple of comment for consideration

internal/controller/common/helper.go
mayankshah1607: good catch, updated

internal/controller/common/helper.go
mayankshah1607: makes sense, updated

bundle/manifests/everest-operator.clusterserviceversion.yaml
```diff
@@ -78,7 +78,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2025-02-14T15:05:36Z"
+    createdAt: "2025-03-10T11:36:03Z"
```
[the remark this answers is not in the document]
mayankshah1607: done

internal/controller/databasecluster_controller.go
```diff
@@ -180,10 +180,10 @@ func (r *DatabaseClusterReconciler) reconcileDB(
...
 	// Running the applier can possibly also mutate the DatabaseCluster,
 	// so we should make sure we push those changes to the API server.
-	updatedDB := db.DeepCopy()
-	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, db, func() error {
-		db.ObjectMeta = updatedDB.ObjectMeta
-		db.Spec = updatedDB.Spec
+	dbCopy := db.DeepCopy()
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, dbCopy, func() error {
+		dbCopy.ObjectMeta = db.ObjectMeta
+		dbCopy.Spec = db.Spec
```
recharte, line 186: I don't fully understand why we are using a copy. What's the reason for this?

tests/e2e/core/pxc/65-assert.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kuttl.dev/v1
+kind: TestAssert
+timeout: 600
+---
+apiVersion: everest.percona.com/v1alpha1
+kind: DatabaseCluster
+metadata:
+  name: test-pxc-cluster
+status:
+  status: ready
```
recharte, line 10: Why isn't this part of `64-assert.yaml`?

api/v1alpha1/databasecluster_types.go
```diff
@@ -122,6 +124,11 @@ type Storage struct {
 	Size resource.Quantity `json:"size"`
 	// Class is the storage class to use for the persistent volume claim
 	Class *string `json:"class,omitempty"`
+	// DisableVolumeExpansion controls whether the storage size of an existing cluster can be increased.
+	// When set to true, the storage size cannot be modified after creation.
+	// By default, this is false, allowing the storage size to be increased by specifying a new size value.
+	// Note: Volume expansion requires support from the underlying storage class.
+	DisableVolumeExpansion bool `json:"disableVolumeExpansion,omitempty"`
```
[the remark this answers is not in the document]
mayankshah1607: It just acts as an additional safety check that can be configured at the DB level:

Similar to how the `enableVolumeExpansion` flag acts as a safety check in the upstream CRs: https://github.com/percona/percona-xtradb-cluster-operator/blob/main/pkg/controller/pxc/volumes.go#L241 (except the default behaviour is reversed)...

internal/controller/databasecluster_controller.go
mayankshah1607: When you use `CreateOrUpdate`:
- it first fetches the current state from Kubernetes and overwrites the object passed to it
- In our case, `db` contains mutations from previous reconciliation steps (like status updates)
- If we passed `db `directly, these mutations would be lost when `CreateOrUpdate` fetches the current state

So the flow of operations would be like so:
Without copy:
1. We mutate db in previous steps
2. Pass `db` to `CreateOrUpdate`
3. `CreateOrUpdate `fetches current state -> overwrites our mutations
4. Our mutations are lost

With copy:
1. We mutate db in previous steps
2. Create `dbCopy` to preserve mutations
3. `CreateOrUpdate` fetches current state -> overwrites `dbCopy`
4. We reapply our mutations in the anonymous function
5. Original mutations are preserved

The only thing different in this PR is that instead of passing `db` directly to `CreateOrUpdate`, we are passing the copy `dbCopy`. This is done so that the mutations done to the status of `db` are not lost (in the code after this, we are updating the status using `db`)

tests/e2e/core/pxc/65-assert.yaml
mayankshah1607: We want to first check that it gets to `resizingVolumes` state, which we do in `64-assert.yaml`. Next, we want to check that resizing was complete and DB is in `ready` state, which is done in `65-assert.yaml`.

Can a single assert step sequentially check it the DB is first in `resizingVolumes` and then `ready`? I was under the impression that for an assertion to pass, ALL the conditions have to be met (which would fail here because the DB cannot be in both `resizingVolumes` and `ready` at once) 🤔

recharte: Ahhh nevermind, I forgot about the `resizingVolumes` state, I assumed that the only thing you were checking in `64-assert.yaml` was the new size.

internal/controller/databasecluster_controller.go
recharte: Thanks for the explanation :hugs:

api/v1alpha1/databasecluster_types.go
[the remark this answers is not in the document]
recharte: :thinking: I don't think we have any plans to add this to the UI and it seems it's an extra field that will not be used in 99.9% of the cases. I'm not sure it's worth having it. I don't oppose having it but I don't see it being useful for users that's why I asked about the use case.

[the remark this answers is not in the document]
mayankshah1607: Okay, I've removed it

recharte (approved):

PR #698, opened 2025-03-13
EVEREST-1877 | PSMDB Volume expansion

mayankshah1607:


internal/controller/providers/psmdb/provider.go
```diff
@@ -147,6 +154,32 @@ func (p *Provider) Status(ctx context.Context) (everestv1alpha1.DatabaseClusterS
...
+			return false, err
+		}
+		for _, sts := range stsList.Items {
+			annots := sts.GetAnnotations()
+			_, ok := annots[psmdbv1.AnnotationPVCResizeInProgress]
+			if ok {
+				return true, nil
+			}
+		}
+		return false, nil
```
oksana-grishchenko, line 178: just a nitpick - looks like this return can be omitted

internal/controller/providers/psmdb/applier.go
```diff
@@ -694,3 +694,42 @@ func (p *applier) genPSMDBBackupSpec() (psmdbv1.BackupSpec, error) {
...
+func configureStorage(
+	ctx context.Context,
+	c client.Client,
+	desired *psmdbv1.ReplsetSpec,
+	current *psmdbv1.ReplsetSpec,
+	db *everestv1alpha1.DatabaseCluster,
+) error {
+	var currentSize resource.Quantity
+	if current != nil {
+		currentSize = current.VolumeSpec.PersistentVolumeClaim.PersistentVolumeClaimSpec.Resources.Requests[corev1.ResourceStorage]
```
oksana-grishchenko, line 716: `.VolumeSpec.PersistentVolumeClaim.PersistentVolumeClaimSpec` technically can be null. not sure what is the chance, but the spec allows it. should we handle that case?

internal/controller/providers/psmdb/provider.go
mayankshah1607: Leftovers from some refactoring, updated!

internal/controller/providers/psmdb/applier.go
mayankshah1607: The `setStorageSize` lamda function below is the only place we are setting the VolumeSpec field, and there we always initialize the `.VolumeSpec.PersistentVolumeClaim.PersistentVolumeClaimSpec` object. So at least right now, we will not get the nil-ptr deref.

But it is still good to add the additional nil-checks for future proofing. Updated, thanks :)

PR #699, opened 2025-03-14
EVEREST-107 Update VictoriaMetrics operator to v0.54.1

recharte:
**Update VictoriaMetrics operator to v0.54.1**
---
**Problem:**
EVEREST-107

**Related pull requests**

- https://github.com/percona/percona-helm-charts/pull/500


go.sum
```diff
@@ -3359,128 +1056,72 @@ honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc/go.mod h1:rf3lG4BRIbNafJWh
 honnef.co/go/tools v0.0.1-2019.2.3/go.mod h1:a3bituU0lyd329TUQxRnasdCoJDkEUEAqEt0JzvZhAg=
 honnef.co/go/tools v0.0.1-2020.1.3/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9vFzvIQ3k=
 honnef.co/go/tools v0.0.1-2020.1.4/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9vFzvIQ3k=
-honnef.co/go/tools v0.1.3/go.mod h1:NgwopIslSNH47DimFoV78dnkksY2EFtX0ajyb3K/las=
-k8s.io/api v0.31.1 h1:Xe1hX/fPW3PXYYv8BlozYqw63ytA92snr96zMW9gWTU=
-k8s.io/api v0.31.1/go.mod h1:sbN1g6eY6XVLeqNsZGLnI5FwVseTrZX7Fv3O26rhAaI=
-k8s.io/apiextensions-apiserver v0.31.1 h1:L+hwULvXx+nvTYX/MKM3kKMZyei+UiSXQWciX/N6E40=
-k8s.io/apiextensions-apiserver v0.31.1/go.mod h1:tWMPR3sgW+jsl2xm9v7lAyRF1rYEK71i9G5dRtkknoQ=
+k8s.io/api v0.18.2/go.mod h1:SJCWI7OLzhZSvbY7U8zwNl9UA4o1fizoug34OV/2r78=
+k8s.io/api v0.18.4/go.mod h1:lOIQAKYgai1+vz9J7YcDZwC26Z0zQewYOGWdyIPUUQ4=
```
oksana-grishchenko, line 1060: I'm wondering if we still should use the replacement
```
k8s.io/api => k8s.io/api v0.31.1
```
if we still have this older v0.18.2 dependency in go.sum🤔

same for the rest of the replacements

internal/controller/monitoringconfig_controller.go
```diff
@@ -153,17 +153,21 @@ func (r *MonitoringConfigReconciler) reconcileVMAgent(ctx context.Context) error
...
-		ExtraArgs: map[string]string{
-			"memory.allowedPercent": "40",
-		},
-		Resources: corev1.ResourceRequirements{
-			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("250m"),
-				corev1.ResourceMemory: resource.MustParse("350Mi"),
+		CommonApplicationDeploymentParams: vmv1beta1.CommonApplicationDeploymentParams{
+			ExtraArgs: map[string]string{
+				"memory.allowedPercent": "40",
```
oksana-grishchenko, line 158: Just curious why 40%? Is that recommended somewhere?

recharte: I have no idea :sweat_smile: I didn't want to change it.

go.sum
recharte: I don't think we need them. AFAIK the only reason we had the replacements was to be able to compile with the old VM operator. Now, this new version is compatible with the versions of the deps we use.

internal/controller/monitoringconfig_controller.go
oksana-grishchenko: Oh, I see, it was like this before😆 For some reason I thought it's a new parameter, that's why I got confused why this exact value is used. I see it was like this since [pmm dbaas](https://github.com/percona/dbaas-controller/blob/9900837a95c73b73eec61bd3b0198a106d2855c6/service/k8sclient/k8sclient.go#L1898-L1900). Never mind then.

oksana-grishchenko (approved):

PR #700, opened 2025-03-17
EVEREST-1927 | volume expansion for PG

mayankshah1607:


internal/controller/databasecluster_controller.go
```diff
@@ -663,6 +669,15 @@ func (r *DatabaseClusterReconciler) initWatchers(controller *builder.Builder) {
...
 	)
 
+	// Since PerconaPGCluster does not expose any info about volume resizing,
+	// we need to directly watch the PostgresCluster objects to track the status.
+	// See: https://perconadev.atlassian.net/browse/K8SPG-748
+	// TODO: Remove this once K8SPG-748 is addressed.
+	controller.Watches(
+		&crunchyv1beta1.PostgresCluster{},
+		&handler.EnqueueRequestForObject{},
+	)
```
oksana-grishchenko, line 679: So the reconciliation will be triggered only for the DB cluster with the same Name and Namespace as the `postgresclusters.postgres-operator.crunchydata.com`. But there is still this duplicated reconciliation bc the controller also watches `perconapgclusters.pgv2.percona.com`, so I'm wondering if we can decrease the number of reconciliation calls by adding some predicate. Looks like excluding the Create, Delete and Generic events won't bring any significant profit, but I might miss something so leaving this comment for consideration.

oksana-grishchenko (approved):
A bit concerned about the overhead in the additional watcher, the checking of the chunchy PG cluster and the pvc, however I don't see more efficient options how we could currently handle the resizing status. Left one comment for consideration just in case it would invoke some thoughts

internal/controller/databasecluster_controller.go
mayankshah1607: I think this is a fair point. Normally we get notified from the `PerconaPGCluster` object, but this watcher is needed only for the PVC resizing status. So it makes sense to have an event filter here to avoid the duplicate reconciliations..

mayankshah1607: FWIW, the default workqueue implementation actually de-duplicates the requests for us: https://github.com/kubernetes/client-go/blob/master/util/workqueue/queue.go#L241

But adding a predicate here is also low effort, so I don't see the harm

mayankshah1607: Updated, thank you for the suggestion

PR #704, opened 2025-03-19
EVEREST-1930 Add upgrading state

recharte:
**Add upgrading state**
---
**Problem:**
EVEREST-1930

**Related pull requests**

- https://github.com/percona/everest/pull/1235
- https://github.com/percona/percona-helm-charts/pull/502

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/internal/controller/providers/pg/provider.go
+++ b/internal/controller/providers/pg/provider.go
@@ -90,6 +90,36 @@ func (p *Provider) Apply(ctx context.Context) everestv1alpha1.Applier {
 	}
 }
 
+// isDatabaseUpgrading returns true if the database is upgrading.
+func (p *Provider) isDatabaseUpgrading(ctx context.Context) (bool, error) {
+	// Get PG pods to check if an upgrade is pending or in progress.
+	listOpts := &client.ListOptions{
+		Namespace:     p.DB.GetNamespace(),
+		LabelSelector: labels.SelectorFromSet(labels.Set{"app.kubernetes.io/component": "pg", "app.kubernetes.io/instance": p.DB.GetName()}),
+	}
+
+	podList := &corev1.PodList{}
+	if err := p.C.List(ctx, podList, listOpts); err != nil {
+		return false, err
+	}
+
+	for _, pod := range podList.Items {
+		for _, container := range pod.Spec.Containers {
+			if container.Name != "database" {
+				continue
+			}
+
+			// If there's a pod with a different image tag from the one
+			// specified in the CR, an upgrade is pending or in progress.
+			if container.Image != p.PerconaPGCluster.Spec.Image {
+				return true, nil
+			}
+		}
+	}
+
+	return false, nil
+}
+
 // Status builds the DatabaseCluster Status based on the current state of the PerconaPGCluster.
 func (p *Provider) Status(ctx context.Context) (everestv1alpha1.DatabaseClusterStatus, error) {
 	c := p.C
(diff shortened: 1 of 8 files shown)


oksana-grishchenko (approved):

internal/controller/providers/pg/provider.go
```diff
@@ -109,6 +111,29 @@ func (p *Provider) Status(ctx context.Context) (everestv1alpha1.DatabaseClusterS
...
+	for _, pod := range podList.Items {
+		for _, container := range pod.Spec.Containers {
+			if container.Name != "database" {
+				continue
+			}
+
+			// If there's a pod with a different image tag from the one
+			// specified in the CR, an upgrade is pending or in progress.
+			if container.Image != pg.Spec.Image {
+				status.Status = everestv1alpha1.AppStateUpgrading
```
maxkondr, line 132: it seems it will be enough to get the first tags mismatch in order to make a decision that DB cluster is in `upgrading` state. No need to check the rest of pods (especially in big clusters)

maxkondr: and I would move this logic to a separate function like `isDatabseClusterInUpgradeState` or similar. the same as we have above `common.IsDatabaseClusterRestoreRunning`

internal/controller/providers/pg/provider.go
```diff
@@ -109,6 +111,29 @@ func (p *Provider) Status(ctx context.Context) (everestv1alpha1.DatabaseClusterS
 		status.Status = everestv1alpha1.AppStateRestoring
 	}
 
+	// Get PG pods to check if an upgrade is pending or in progress.
+	listOpts := &client.ListOptions{
+		Namespace:     p.DB.GetNamespace(),
+		LabelSelector: labels.SelectorFromSet(labels.Set{"app.kubernetes.io/component": "pg", "app.kubernetes.io/instance": p.DB.GetName()}),
+	}
+	podList := &corev1.PodList{}
+	if err := c.List(ctx, podList, listOpts); err != nil {
```
maxkondr, line 120: One thing that concerns me - we will check the pods almost on each reconciliation iteration. Are there any other evidences of DB cluster upgrade process (maybe in pg-operator's DB cluster resource)?

internal/controller/providers/psmdb/provider.go
```diff
@@ -139,6 +139,12 @@ func (p *Provider) Status(ctx context.Context) (everestv1alpha1.DatabaseClusterS
 		status.Status = everestv1alpha1.AppStateRestoring
 	}
 
+	// If the current version of the database is different from the version in
+	// the CR, an upgrade is pending or in progress.
+	if p.DB.Spec.Engine.Version != "" && psmdb.Status.MongoVersion != "" && p.DB.Spec.Engine.Version != psmdb.Status.MongoVersion {
```
maxkondr, line 144: it really looks like a better approach move this logic to DB operators and reflect the `upgrading` state in db-operator's resource, isn't it?

internal/controller/providers/pg/provider.go
recharte: Unfortunately not, the PXC and PSMDB operators do have the version in their CRs but there's no such field in the PG CR :disappointed: I thought about opening a feature request for the cloud team to add a similar field.

internal/controller/providers/psmdb/provider.go
recharte: Yes, I also thought about that and was thinking about discussing this possibility with the cloud team in our next sync.

internal/controller/providers/pg/provider.go
recharte: Agree, done

maxkondr (approved):

PR #729, opened 2025-04-09
EVEREST-1987 | EVEREST-1700 Update PXCO to v1.17.0 and fix restart issue when monitoring is enabled

recharte:
**Update PXCO to v1.17.0 and fix restart issue when monitoring is enabled**
---
**Problem:**
EVEREST-1987
EVEREST-1700

This PR introduces:
* The upgrade of PXCO to v1.17.0
* A workaround for K8SPXC-1568 that affects v1.17.0, see [2a9b4a1](https://github.com/percona/everest-operator/pull/729/commits/2a9b4a1eb599ae296490c79c866fd608f216de6b) and [f8aa1b3](https://github.com/percona/everest-operator/pull/729/commits/f8aa1b3b9100d5354dda5d2cd03aed216275ec38)
* A fix for EVEREST-1700, see [1fad7b9](https://github.com/percona/everest-operator/pull/729/commits/1fad7b9fe491aadd140ebc57809e57e43a31301d)
* A workaround for broken PITR process, see: https://github.com/percona/everest-operator/pull/739

**Related pull requests**

- https://github.com/percona/everest-catalog/pull/58
- https://github.com/percona/everest/pull/1315

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**


oksana-grishchenko (approved):

mayankshah1607 (approved):

recharte:
@mayankshah1607 and @oksana-grishchenko please re-review the PR because I introduced:
* A workaround for K8SPXC-1568 that affects v1.17.0, see [2a9b4a1](https://github.com/percona/everest-operator/pull/729/commits/2a9b4a1eb599ae296490c79c866fd608f216de6b) and [f8aa1b3](https://github.com/percona/everest-operator/pull/729/commits/f8aa1b3b9100d5354dda5d2cd03aed216275ec38)
* A fix for EVEREST-1700, see [1fad7b9](https://github.com/percona/everest-operator/pull/729/commits/1fad7b9fe491aadd140ebc57809e57e43a31301d)

internal/controller/providers/pxc/applier.go
```diff
@@ -117,6 +173,20 @@ func (p *applier) Engine() error {
...
+	// cluster stuck in the 'initializing' state.
+	// See: https://perconadev.atlassian.net/browse/K8SPXC-1568
+	// To work around this, we need to set the password ourselves to bypass the
+	// buggy auto generation.
+	if p.DBEngine.Status.OperatorVersion == "1.17.0" {
+		err := p.ensureSecretHasProxyAdminPassword(pxc.Spec.SecretsName)
+		if err != nil {
+			return fmt.Errorf("failed to set proxyadmin password: %w", err)
+		}
+	}
```
oksana-grishchenko, line 188: Just wanted to highlight that with this workaround:
- we assume that in 1.18.0 the problem will be fixed. So far the fix is indeed planned for 1.18.0. 
- we accept that the user can't change this password. Not that Everest was explicitly allowing that before but technically there was still an option to change the password in secrets. With this workaround the change won't take any effect.

This is a non-blocking comment, feel free to resolve it.

oksana-grishchenko (approved):

internal/controller/providers/pxc/applier.go
```diff
@@ -117,6 +143,39 @@ func (p *applier) Engine() error {
...
+		secret := &corev1.Secret{}
+		err := p.C.Get(p.ctx, types.NamespacedName{
+			Name:      p.DB.Spec.Engine.UserSecretsName,
+			Namespace: p.DB.GetNamespace(),
+		}, secret)
+		if client.IgnoreNotFound(err) != nil {
+			return err
+		}
+
+		if _, ok := secret.Data["proxyadmin"]; !ok {
```
mayankshah1607: Is it enough to check just for the existence of the `proxyadmin` key?

What I mean to ask is - are we sure that the pxc-operator won't set this key (with the buggy password) before everest-operator? With the current logic, if the pxc-operator sets a buggy password (before this code executes), then everest-operator will not do anything because the key exists in the secret (and that's all we check)

internal/controller/providers/pxc/applier.go
```diff
@@ -102,6 +109,55 @@ func configureStorage(
...
+		Name:      p.DB.Spec.Engine.UserSecretsName,
+		Namespace: p.DB.GetNamespace(),
+	}, secret)
+	if client.IgnoreNotFound(err) != nil {
+		return err
+	}
+
+	if _, ok := secret.Data["proxyadmin"]; ok {
+		return nil
+	}
```
mayankshah1607, line 143: Is it possible for the pxc-operator to set this password before everest-operator? In that case, the key would still exist with the buggy password, and we would allow it because we're just checking for the existence for the key

recharte: I'm pretty sure it's guaranteed that this code runs before creating the PXC CR in the first reconciliation so I don't think that there's a chance for the PXC operator to mess with this secret.

internal/controller/providers/pxc/applier.go
recharte: I confirmed with Slava that this will be fixed in 1.18.0.
Why wouldn't the user be able to change this password? We are only generating a password if it doesn't exist in the secret already. We will not forbid a password change nor will we replace an existing one. Am I missing something?

oksana-grishchenko: I got under impression that with `CreateOrUpdateSecretData` we're replacing the existing password which we actually don't. I missed the part where we check wether the password is already there and do nothing in that case. So never mind the second note, sorry for the confusion.

mayankshah1607:
@recharte just FYI for when you're back, the PITR broke with PXC 1.17.0, so we had to create another workaround. See PR: https://github.com/percona/everest-operator/pull/739

PR #733, opened 2025-04-11
EVEREST-1299 | [docs] add basic guide on using the CRDs

mayankshah1607:
Since there has been a lot of interest in the community for using Everest using CRDs (for e.g, using GitOps integrations), this PR adds some documentation around how to correctly use the CRDs along with a bunch of examples.


docs/guides/crd_guide.md
```diff
@@ -0,0 +1,497 @@
...
+        memory: "100Mi"
+```
+
+## Examples
+
+### Complete PostgreSQL Cluster Example
+
+```yaml
+apiVersion: everest.percona.com/v1alpha1
+kind: DatabaseCluster
```
oksana-grishchenko, line 341: Would it be too bad if we provide links to the yamls instead of the plain-text yamls? I'm just a bit concerned about supporting duplicated yamls in two places😅

docs/guides/crd_guide.md
```diff
@@ -0,0 +1,497 @@
...
+kind: DatabaseClusterRestore
+metadata:
+  name: latest-pitr
+spec:
+  dbClusterName: my-database
+  dataSource:
+    dbClusterBackupName: base-backup
+    pitr:
+      type: latest
+```
```
oksana-grishchenko, line 274: although the CRs allow the `latest` restoration type I'm not sure if our QA ever touched that functionality and we never officially supported it. Should we still mention it in this guide?

oksana-grishchenko (approved):

docs/guides/crd_guide.md
oksana-grishchenko: Let's have some third opinion here then, I'm not sure what if we should remove it or not... @recharte wdyt?

mayankshah1607: > we never officially supported it

technically we also never officially supported using CRDs but here we are 😂 (jokes aside, yeah, lets see what Diogo has to say)

mayankshah1607: since Diogo is on vacation, I'm removing this example so we can resolve and merge.. We can add back later if needed

PR #737, opened 2025-04-17
EVEREST-1520 Implement PodSchedulingPolicy feature

maxkondr:
EVEREST-1520 Add PodSchedulingPolicy feature support to Everest

**Related pull requests**
Everest PR - https://github.com/percona/everest/pull/1322
Helm chart PR - https://github.com/percona/percona-helm-charts/pull/531


api/v1alpha1/podschedulingpolicy_types.go
```diff
@@ -0,0 +1,102 @@
...
+	AffinityConfig *AffinityConfig `json:"affinityConfig,omitempty"`
+}
+
+// PodSchedulingPolicyStatus defines the observed state of PodSchedulingPolicy.
+type PodSchedulingPolicyStatus struct {
+	// Used is a flag that indicates if the policy is used by any DB cluster.
+	// +kubebuilder:default=false
+	Used bool `json:"used,omitempty"`
+	// ObservedGeneration is the most recent generation observed for this PodSchedulingPolicy.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
mayankshah1607, line 71:
```suggestion
	// InUse is a flag that indicates if the policy is used by any DB cluster.
	// +kubebuilder:default=false
	InUse bool `json:"used,omitempty"`
	// LastObservedGeneration is the last generation observed for this object by the controller.
	LastObservedGeneration int64 `json:"observedGeneration,omitempty"`
```

internal/controller/podschedulingpolicy_controller.go
```diff
@@ -0,0 +1,211 @@
...
+		logger.Error(err, fmt.Sprintf("Failed to fetch DB clusters that use pod scheduling policy='%s'", psp.GetName()))
+		return ctrl.Result{}, errors.Join(err, fmt.Errorf("failed to fetch DB clusters that use pod scheduling policy='%s': %w", psp.GetName(), err))
+	}
+
+	// Update the status and finalizers of the PodSchedulingPolicy object after the reconciliation.
+	defer func() {
+		// Nothing to process on delete events
+		if !psp.GetDeletionTimestamp().IsZero() {
+			return
+		}
```
mayankshah1607, line 81: Why do we do this? It means that if there is a deletion timestamp, we will not update the status

internal/controller/podschedulingpolicy_controller.go
```diff
@@ -0,0 +1,211 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *PodSchedulingPolicyReconciler) ensureFinalizers(ctx context.Context, used bool, psp *everestv1alpha1.PodSchedulingPolicy) error {
+	currentFinalizers := psp.GetFinalizers()
+	var desiredFinalizers []string
+	if used {
+		desiredFinalizers = append(currentFinalizers, everestv1alpha1.UsedResourceFinalizer)
+		// remove duplicates
+		desiredFinalizers = slices.Compact(desiredFinalizers)
```
mayankshah1607, line 106: You can use `controllerutil.AddFinalizer` and `controllerutil.RemoveFinalizer` instead since it takes care of the duplicates

internal/controller/podschedulingpolicy_controller.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if err = r.ensureFinalizers(ctx, len(dbList.Items) > 0, psp); err != nil {
+		logger.Error(err, fmt.Sprintf("failed to update finalizers for pod scheduling policy='%s'", psp.GetName()))
+		return ctrl.Result{}, errors.Join(err, fmt.Errorf("failed to update finalizers for pod scheduling policy='%s': %w", psp.GetName(), err))
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *PodSchedulingPolicyReconciler) ensureFinalizers(ctx context.Context, used bool, psp *everestv1alpha1.PodSchedulingPolicy) error {
```
mayankshah1607, line 100: Please add a comment describing how the finalizers are eventually removed for uninstallation of Everest

internal/controller/podschedulingpolicy_controller.go
```diff
@@ -0,0 +1,211 @@
...
+	pspEventsPredicate := predicate.Funcs{
+		// Nothing to process on create events
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
+		},
+
+		// Allow update events only if the policy is in the system namespace.
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return e.ObjectOld.GetNamespace() == common.SystemNamespace
+		},
```
mayankshah1607, line 165: Shouldn't we filter for all types of events, and not just for UPDATE? For example, if we are creating a PSP in some namespace `test`, we will be processing that event. Shouldn't that be skipped as well?

mayankshah1607 (changes requested):
Reviewed partially, will continue again after a while

api/v1alpha1/podschedulingpolicy_types.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=psp
+// +kubebuilder:selectablefield:JSONPathstring=".spec.engineType"
+// +kubebuilder:printcolumn:name="Engine",type="string",JSONPath=".spec.engineType",description="DB engine type the policy can be applied to"
+// +kubebuilder:printcolumn:name="Used",type="string",JSONPath=".status.used",description="Indicates if the policy is used by any DB cluster"
```
recharte, line 79: Didn't we agree that the `PodSchedulingPolicy` would be cluster scoped CRD?
Kubebuilder has the following annotation to set a CRD's scope:
```
// +kubebuilder:resource:scope=Cluster
```

recharte: This also means that when creating a `PodSchedulingPolicy` CR you won't need to store it in the `everest-system` namespace as is currently being done.

internal/controller/podschedulingpolicy_controller.go
maxkondr: Is it really required to update the state of resource that is being deleted anyway? I just think this is not needed at all.

api/v1alpha1/podschedulingpolicy_types.go
recharte: Will this impose any limitations when we add the ability to customize the system's namespace (https://github.com/percona/everest/issues/1159)?

internal/controller/common/helper.go
```diff
@@ -573,6 +573,36 @@ func GetDBMonitoringConfig(
 	return monitoring, nil
 }
 
+// GetPodSchedulingPolicy returns the PodSchedulingPolicy object by name.
+func GetPodSchedulingPolicy(ctx context.Context, c client.Client, pspName string) (*everestv1alpha1.PodSchedulingPolicy, error) {
+	psp := &everestv1alpha1.PodSchedulingPolicy{}
+	if pspName != "" {
+		if err := c.Get(ctx, types.NamespacedName{
+			Namespace: SystemNamespace,
```
recharte, line 581: To avoid re-introducing this hard-coded namespace and potentially making it harder to fix https://github.com/percona/everest/issues/1159 do you think it would be okay to use the namespace where the operator is running instead?

internal/controller/databasecluster_controller.go
```diff
@@ -792,19 +873,18 @@ func (r *DatabaseClusterReconciler) databaseClustersInObjectNamespace(ctx contex
...
+// 	keyPath, namespace, name string,
+// ) (*everestv1alpha1.DatabaseClusterList, error) {
+// 	attachedDatabaseClusters := &everestv1alpha1.DatabaseClusterList{}
+// 	listOps := &client.ListOptions{
+// 		FieldSelector: fields.OneTermEqualSelector(keyPath, name),
+// 		Namespace:     namespace,
+// 	}
+// 	err := r.List(ctx, attachedDatabaseClusters, listOps)
+// 	return attachedDatabaseClusters, err
+// }
```
recharte, line 887: Since this is no longer used, let's delete it.

internal/controller/providers/pg/applier.go
```diff
@@ -216,15 +204,15 @@ func (p *applier) Proxy() error {
...
-	}
+	// pg.Spec.Proxy.PGBouncer.Affinity = common.DefaultAffinitySettings().DeepCopy()
+	// // New affinity settings (added in 1.2.0) must be applied only when PG is upgraded to 2.4.1.
+	// // This is a temporary workaround to make sure we can make this change without an automatic restart.
+	// // TODO: fix this once https://perconadev.atlassian.net/browse/EVEREST-1413 is addressed.
+	// crVersion := goversion.Must(goversion.NewVersion(pg.Spec.CRVersion))
+	// if p.DB.Status.Status != everestv1alpha1.AppStateNew &&
+	// 	crVersion.LessThan(goversion.Must(goversion.NewVersion("2.4.1"))) {
+	// 	pg.Spec.Proxy.PGBouncer.Affinity = p.currentPGSpec.Proxy.PGBouncer.Affinity
+	// }
```
recharte, line 215:
```suggestion
```

internal/controller/podschedulingpolicy_controller.go
mayankshah1607: > Is it really required to update the state of resource that is being deleted anyway?
Some CRDs have a `deleting` state, so if there is additional clean-up logic needed, you would need to reconcile a deleting object. But in this case, we don't have such a requirement, so I think it is fine. I just wanted to ensure that this was intentional

internal/controller/podschedulingpolicy_controller.go
```diff
@@ -0,0 +1,211 @@
...
+
+	pspEventsPredicate := predicate.Funcs{
+		// Nothing to process on create events
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
+		},
+
+		// Allow update events only if the policy is in the system namespace.
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return e.ObjectOld.GetNamespace() == common.SystemNamespace
```
mayankshah1607, line 164: Same as https://github.com/percona/everest-operator/pull/737/files#r2076015522 , I'd suggest using the value of `SystemNamespace` read from `cmd/main.go` rather than hard-coding

internal/controller/databasecluster_controller.go
```diff
@@ -717,7 +753,52 @@ func (r *DatabaseClusterReconciler) initWatchers(controller *builder.Builder) {
...
+				// Nothing to process on delete events
+				DeleteFunc: func(e event.DeleteEvent) bool {
+					return false
+				},
+				// Nothing to process on generic events
+				GenericFunc: func(e event.GenericEvent) bool {
+					return false
+				},
+			},
+		),
```
mayankshah1607, line 801: Same predicate it used for the PSP controller, can we move this to a common function to reduce duplicate code?

internal/controller/providers/pg/applier.go
```diff
@@ -274,6 +262,92 @@ func (p *applier) Monitoring() error {
...
+		// crVersion := goversion.Must(goversion.NewVersion(pg.Spec.CRVersion))
+		// if p.DB.Status.Status == everestv1alpha1.AppStateReady &&
+		// 	crVersion.LessThan(goversion.Must(goversion.NewVersion("2.4.1"))) {
+		//
+		// 	for i := range len(pg.Spec.InstanceSets) {
+		// 		// all instances in sets have the same affinity configuration.
+		// 		pg.Spec.InstanceSets[i].Affinity = p.currentPGSpec.InstanceSets[0].Affinity
+		// 	}
+		// 	pg.Spec.Proxy.PGBouncer.Affinity = p.currentPGSpec.Proxy.PGBouncer.Affinity
+		// }
```
mayankshah1607, line 305: Should this be uncommented?

How are we going to ensure that this new change preserves the existing affinity configuration?

internal/controller/providers/pg/applier.go
```diff
@@ -274,6 +262,92 @@ func (p *applier) Monitoring() error {
...
+
+		return nil
+	}
+
+	var psp *everestv1alpha1.PodSchedulingPolicy
+	var err error
+	psp, err = common.GetPodSchedulingPolicy(p.ctx, p.C, pspName)
```
mayankshah1607, line 312:
```suggestion
	psp, err := common.GetPodSchedulingPolicy(p.ctx, p.C, pspName)
```

internal/controller/providers/psmdb/applier.go
```diff
@@ -343,6 +327,119 @@ func (p *applier) Monitoring() error {
...
+		// 		psmdb.Spec.Sharding.Mongos.MultiAZ.Affinity = p.currentPSMDBSpec.Sharding.Mongos.MultiAZ.Affinity
+		// 	}
+		// }
+
+		return nil
+	}
+
+	var psp *everestv1alpha1.PodSchedulingPolicy
+	var err error
+	psp, err = common.GetPodSchedulingPolicy(p.ctx, p.C, pspName)
```
mayankshah1607, line 388:
```suggestion
	psp, err := common.GetPodSchedulingPolicy(p.ctx, p.C, pspName)
```

internal/controller/providers/pxc/applier.go
```diff
@@ -300,6 +291,105 @@ func (p *applier) Monitoring() error {
...
+		// 	pxc.Spec.ProxySQL.PodSpec.Affinity = p.currentPerconaXtraDBClusterSpec.ProxySQL.PodSpec.Affinity
+		// default:
+		// 	return fmt.Errorf("invalid proxy type %s", p.DB.Spec.Proxy.Type)
+		// }
+		return nil
+	}
+
+	var psp *everestv1alpha1.PodSchedulingPolicy
+	var err error
+	psp, err = common.GetPodSchedulingPolicy(p.ctx, p.C, pspName)
```
mayankshah1607, line 344:
```suggestion
	psp, err := common.GetPodSchedulingPolicy(p.ctx, p.C, pspName)
```

internal/controller/providers/psmdb/applier.go
```diff
@@ -343,6 +327,117 @@ func (p *applier) Monitoring() error {
...
+		// 		// all replsets have the same affinity config.
+		// 		psmdb.Spec.Replsets[i].MultiAZ.Affinity = p.currentPSMDBSpec.Replsets[0].MultiAZ.Affinity
+		// 	}
+		//
+		// 	// check that sharding was enabled in a previous state of the upstream cluster.
+		// 	if p.currentPSMDBSpec.Sharding.Enabled {
+		// 		// Config Server
+		// 		psmdb.Spec.Sharding.ConfigsvrReplSet.MultiAZ.Affinity = p.currentPSMDBSpec.Sharding.ConfigsvrReplSet.MultiAZ.Affinity
+		// 		// Proxy
+		// 		psmdb.Spec.Sharding.Mongos.MultiAZ.Affinity = p.currentPSMDBSpec.Sharding.Mongos.MultiAZ.Affinity
```
oksana-grishchenko, line 379: Note: in theory, `p.currentPSMDBSpec.Sharding.ConfigsvrReplSet` and `p.currentPSMDBSpec.Sharding.Mongos` can be nil

internal/controller/providers/pxc/applier.go
```diff
@@ -300,6 +291,103 @@ func (p *applier) Monitoring() error {
...
+		// See: https://perconadev.atlassian.net/browse/EVEREST-1413
+		// if p.DB.Status.Status == everestv1alpha1.AppStateReady {
+		// 	pxc.Spec.PXC.PodSpec.Affinity = p.currentPerconaXtraDBClusterSpec.PXC.Affinity
+		// }
+		// // Changing proxy type is not supported, so we copy affinity rules from the previous state of the same proxy type
+		// switch p.DB.Spec.Proxy.Type {
+		// case everestv1alpha1.ProxyTypeHAProxy:
+		// 	pxc.Spec.HAProxy.PodSpec.Affinity = p.currentPerconaXtraDBClusterSpec.HAProxy.PodSpec.Affinity
+		// case everestv1alpha1.ProxyTypeProxySQL:
+		// 	pxc.Spec.ProxySQL.PodSpec.Affinity = p.currentPerconaXtraDBClusterSpec.ProxySQL.PodSpec.Affinity
```
oksana-grishchenko, line 335: Note: in theory, `p.currentPerconaXtraDBClusterSpec.HAProxy` and `p.currentPerconaXtraDBClusterSpec.ProxySQL` can be nil

api/v1alpha1/podschedulingpolicy_types.go
maxkondr: now PSP controller uses system namespace name from config

internal/controller/podschedulingpolicy_controller.go
```diff
@@ -0,0 +1,192 @@
...
+		if !psp.GetDeletionTimestamp().IsZero() {
+			return
+		}
+
+		psp.Status.InUse = len(dbList.Items) > 0
+		psp.Status.LastObservedGeneration = psp.GetGeneration()
+		if err = r.Client.Status().Update(ctx, psp); err != nil {
+			rr = ctrl.Result{}
+			logger.Error(err, fmt.Sprintf("failed to update status for pod scheduling policy='%s'", psp.GetName()))
+			rerr = errors.Join(err, fmt.Errorf("failed to update status for pod scheduling policy='%s': %w", psp.GetName(), err))
```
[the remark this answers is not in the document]
maxkondr: Done

internal/controller/providers/psmdb/applier.go
maxkondr: Yes, I know.
the important thing here is the following:
`applier.PodSchedulingPolicy()` shall be called after `applier.Engine()` and `applier.Proxy()` because exactly in those functions `spec..Sharding.ConfigsvrReplSet` and `Spec.Sharding.Mongos` are initialised

internal/controller/providers/pxc/applier.go
maxkondr: the same as for PSMDB

internal/controller/providers/pg/applier.go
recharte: Like we discussed today, since we enforce only 2 supported operator versions per everest version and in v1.6.0 we only supported PGO v2.5.0 and v2.6.0, there can't be any PG DBs in v1.6.0 that are still using a CR version older than v2.4.1. Therefore, this is essentially dead code and can be safely removed.

api/v1alpha1/podschedulingpolicy_types.go
recharte: Sorry to bring this up again but after getting involved in [this discussion](https://github.com/percona/everest/pull/1322#discussion_r2075408444) I don't feel 100% convinced that we are better off with a namespace scope for this resource.

> User/Admin doesn't know where exactly they are located, for them nothing is changed in general.

This is true for UI/API users but for operator/CR users they still need to know that all `PodSchedulingPolicy` CRs need to be created in the same namespace as the operator, which isn't obvious. Users might even think that since their DBCs reference a `PodSchedulingPolicy` that this policy should exist in the same namespace, which isn't true.

> it is more handy to allocate all Everest's resources in system namespace and do not spread them among the whole K8S cluster -> less mess in cluster.

I don't consider it messy to have cluster scope resources. The only benefit is that deleting the namespace would delete all the `PodSchedulingPolicy` CRs and if you have them cluster scoped you'd need some cleanup when uninstalling.

In general, I think that forcing this resource to be namespaced scope when we treat it as a global resource from the UI/API/RBAC standpoint can be confusing. The only real drawback I see is that with a cluster scope resource the cleanup on uninstallation wouldn't be automatically done by deleting the namespace :shrug: 

@mayankshah1607 @oksana-grishchenko please share your thoughts on this topic.

oksana-grishchenko: It is indeed confusing that in the RBAC we define the `psp` as a namespace-scoped resource (which in fact it is right now). However I'm considering what could go wrong if we make psps cluster-wide. For example, in the psp we can mention the specific labels. Should the information about labels in everest-managed namespaces be available cluster-wide? Rather not, because the labels and their values can reveal internal information cluster-wide which is a security concern.

recharte: Which labels are you referring to? :thinking: 

Nonetheless, it's not true that by making a resource cluster scoped suddenly allows anyone in the cluster to access it.
Each ServiceAccount will still need to have a ClusterRoleBinding to a ClusterRole that gives permission to this particular resource type. So it really isn't any different from the RoleBinding to the Role that would need to be in place for a given ServiceAccount to have access to a namespaced resource in a given namespace.

Since we don't expect some k8s users to have access to a small subset of PodSchedulingPolicies, I don't see any practical difference in terms of security here.

We would only benefit from this resource being namespaced if we intended it to exist in multiple namespaces with potentially different k8s RBAC roles binded to SAs, limiting access to only a subset of them.

mayankshah1607: Logically it makes sense to define these as cluster-scoped resources because such a type represents behaviour rather than workloads that need isolation. For example, the now discontinued k8s type `PodSecurityPolicy` (the original PSP) was also a cluster-scoped resource. 

I'm in agreement with all the points made by Diogo above, especially:

> I think that forcing this resource to be namespaced scope when we treat it as a global resource from the UI/API/RBAC standpoint can be confusing

I don't see any real benefit in keeping them namespaced in k8s and treating them as global resources in Everest. If anything, it just creates confusion, so I'd opt for making them cluster-scoped. The only issue that may arise is that we should not expose any DB related info on them (for e.g, via labels) otherwise we have to rely on users configuring correct RBAC. But I don't think we're doing this, so its okay. 

I also don't think there will be any issues during uninstallation because `helm uninstall ...` will delete all resources it creates regardless of their scope.

mayankshah1607: (As an aside.. (and for transparency) Max and I did talk about having PSPs be namespaced objects, but we didn't see any problems then. But IMO - after seeing the implementation and having these discussions, I think we're better off with cluster-scoped only like planned initially)

recharte: >I also don't think there will be any issues during uninstallation because helm uninstall ... will delete all resources it creates regardless of their scope.

We have the defaults that are created by helm but we'll also have user-created policies which won't be automatically delete by helm, right?

mayankshah1607: Right, of course, that skipped my mind.

Yeah, helm won't clean that up. But that's the case with every user-created resource, that's why we have this pre-delete hook for them - https://github.com/percona/percona-helm-charts/blob/main/charts/everest/charts/common/templates/_db_resources_cleanup.yaml.tpl#L77

I guess we'd have to use the same here

maxkondr: Updated to make PSP cluster-scoped

api/v1alpha1/consts.go
```diff
@@ -20,7 +20,7 @@ const (
 	// User cannot delete resources marked by this finalizer directly.
 	ReadOnlyFinalizer = "everest.percona.com/readonly-protection"
 
-	// UsedResourceFinalizer is the finalizer for marking the resource as "in-use"
+	// InUseResourceFinalizer is the finalizer for marking the resource as "in-use"
 	// and prevents deletion of the resource.
-	UsedResourceFinalizer = "everest.percona.com/used-protection"
+	InUseResourceFinalizer = "everest.percona.com/used-protection"
```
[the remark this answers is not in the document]
maxkondr: Done

oksana-grishchenko (approved):

internal/controller/podschedulingpolicy_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	if used {
+		updated = controllerutil.AddFinalizer(psp, everestv1alpha1.InUseResourceFinalizer)
+	} else {
+		updated = controllerutil.RemoveFinalizer(psp, everestv1alpha1.InUseResourceFinalizer)
+	}
+
+	if updated {
+		if err := r.Update(ctx, psp); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
maxkondr: Done

recharte (approved):
Well done Max :muscle:

PR #739, opened 2025-04-22
EVEREST-2011 | Fix PITR in PXC 1.17.0

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-2011

PITR does not work for PXC 1.17.0

**Related pull requests**

- https://github.com/percona/everest-operator/pull/729

**Cause:**

[K8SPXC-1152](https://perconadev.atlassian.net/browse/K8SPXC-1152) introduces some new [internal changes](https://github.com/percona/percona-xtradb-cluster-operator/pull/1610) to how PITR is managed. In addition to handling the `.spec.paused` field during the restore process, the controller now also temporarily changes the size of the engine and proxy components (though the reason for this is still unclear). This causes a conflict between the `everest-operator` and the `pxc-operator`, as they both attempt to control the same set of fields—ultimately leading to a broken PITR process.

**Solution:**

Originally, the pxc-operator only chanaged the `.spec.paused` field during restores, and this was easy to replicate in the everest-operator. However, now as the restore process has grown more complex, replicating that logic is no longer practical. Instead, we now opt to skip reconciliation entirely while a restore is in progress.

To support this, we’ve introduced a new reconciliationHooks interface, which allows providers to execute custom logic at specific points in the reconciliation lifecycle. As part of this, we’ve added a RunPreReconciliationHook method that checks whether a restore is in progress and, if so, skips reconciliation.

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

[K8SPXC-1152]: https://perconadev.atlassian.net/browse/K8SPXC-1152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/controller/providers/pxc/provider.go
+++ b/internal/controller/providers/pxc/provider.go
@@ -236,15 +236,14 @@ func (p *Provider) Status(ctx context.Context) (everestv1alpha1.DatabaseClusterS
 const defaultRestoreRequeueDuration = 15 * time.Second
 
 func (p *Provider) isRestoreInProgress(ctx context.Context) (bool, error) {
-	var pxcRestoreList pxcv1.PerconaXtraDBClusterRestoreList
-	if err := p.C.List(ctx, &pxcRestoreList, client.InNamespace(p.DB.GetNamespace())); err != nil {
+	restores := &everestv1alpha1.DatabaseClusterRestoreList{}
+	if err := p.C.List(ctx, restores, client.InNamespace(p.DB.GetNamespace())); err != nil {
 		return false, err
 	}
-	for _, restore := range pxcRestoreList.Items {
-		if restore.Status.State == pxcv1.RestoreFailed || restore.Status.State == pxcv1.RestoreSucceeded || restore.Spec.PXCCluster != p.DB.GetName() {
-			continue
+	for _, dbr := range restores.Items {
+		if dbr.IsInProgress() && dbr.Spec.DBClusterName == p.DB.GetName() {
+			return true, nil
 		}
-		return true, nil
 	}
 	return false, nil
 }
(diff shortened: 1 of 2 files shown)


internal/controller/providers/pxc/provider.go
```diff
@@ -266,6 +231,40 @@ func (p *Provider) Status(ctx context.Context) (everestv1alpha1.DatabaseClusterS
...
+// after the specified duration.
+const defaultRestoreRequeueDuration = 15 * time.Second
+
+func (p *Provider) isRestoreInProgress(ctx context.Context) (bool, error) {
+	var pxcRestoreList pxcv1.PerconaXtraDBClusterRestoreList
+	if err := p.C.List(ctx, &pxcRestoreList, client.InNamespace(p.DB.GetNamespace())); err != nil {
+		return false, err
+	}
+	for _, restore := range pxcRestoreList.Items {
+		if restore.Status.State == pxcv1.RestoreFailed || restore.Status.State == pxcv1.RestoreSucceeded || restore.Spec.PXCCluster != p.DB.GetName() {
```
oksana-grishchenko, line 244: I'm wondering if we could check the `dbr` resources instead of the `pxc-restore` - for `dbr` we have this `IsComplete()` which would avoid duplicated logic

oksana-grishchenko (approved):
LGTM. One thing that bugs me - please ensure that everest does not allow restoration of paused clusters and if it does, please check if the restoration of paused clusters works correctly with the new pxc version.

internal/controller/providers/pxc/provider.go
mayankshah1607: updated

PR #763, opened 2025-06-04
EVEREST-2094 | Allow specifying list of user secret keys in DBEngine

mayankshah1607:
#### Problem

As a part of the Data importers feature, some importers will require users to fill out the user secret keys. However, there is no information available to the user as to which keys are available and need to be filled out in the secret.

#### Solution

The DBEngine shall contain a list of supported keys that must go into the Secret. The UI shall render these keys in the form. This way, we're able to guide users into filling out the Secret data correctly.

Note - this PR only updates the DBEngine CRD to support specifying such keys. We shall update the actual DBEngine definitions in the Helm chart as a separate PR.


api/v1alpha1/databaseengine_types.go
```diff
@@ -75,6 +75,27 @@ type (
...
+	SecretKeys SecretKeys `json:"secretKeys,omitempty"`
+}
+
+// SecretKeys contains the definition of the various Secrets that
+// the given DBEngine supports.
+type SecretKeys struct {
+	// User secret keys are used to store the details of the users.
+	User []SecretKey `json:"user,omitempty"`
+}
+
```
oksana-grishchenko, line 92: Have you considered to have `[]SecretKey` where each SecretKey contains the type (in this case, the type would be 'user') instead of the `SecretKeys` structure? Do you expect the structure to be different for different kinds of secrets?

mayankshah1607: Please see section 8.4 of this doc - https://www.notion.so/percona/External-Backup-Imports-1ea674d091f38005bd5bc063a9d1862c?source=copy_link#1ed674d091f38050b996c9d8aff34dc8

oksana-grishchenko (approved):

PR #765, opened 2025-06-05
EVEREST-2094 | [FB] Data importers

mayankshah1607:


internal/consts/finalizers.go
```diff
@@ -0,0 +1,41 @@
...
+
+const (
+	EverestFinalizerPrefix = "everest.percona.com/"
+
+	// ForegroundDeletionFinalizer is the finalizer that ensures foreground deletion for the resource.
+	ForegroundDeletionFinalizer = "foregroundDeletion"
+
+	// DBBackupCleanupFinalizer is the finalizer for cleaning up DatabaseClusterBackup.
+	// Deprecated: We keep this for backward compatibility.
+	DBBackupCleanupFinalizer = EverestFinalizerPrefix + "dbb-cleanup"
```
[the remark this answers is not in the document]
mayankshah1607: I think we still need it -- there could be users who have clusters created from older Everest version, so they may still have this finaliser present. We are not using it anywhere else, it is only for preserving backward compatibility

oksana-grishchenko (approved):

.github/workflows/ci.yml
```diff
@@ -249,8 +266,52 @@ jobs:
...
+          curl -s -L -o ${TOOLS_PATH}/kubectl-kuttl https://github.com/kudobuilder/kuttl/releases/download/v0.22.0/kubectl-kuttl_0.22.0_linux_x86_64  
+          chmod +x ${TOOLS_PATH}/kubectl-kuttl
+
+      - name: Kubectl version
+        run: |
+          kubectl version --client --output=yaml
+          echo $(kubectl kuttl --version) installed
+
+      - name: Run data importer tests
+        if: contains(github.event.pull_request.labels.*.name, 'test/e2e') || contains(github.event.pull_request.labels.*.name, 'test/e2e-data-importer')
```
maxkondr, line 312: seems not required as soon as the same condition is defined on Job level

mayankshah1607: updated

PR #771, opened 2025-06-11
EVEREST-2074 | WIP: PerconaPGRestore based DataImporter

mayankshah1607:


internal/data-importer/cmd/pg/import.go
```diff
@@ -0,0 +1,429 @@
...
+	log.Info().Msgf("Successfully imported PostgreSQL data from %s to %s/%s", backupPath, namespace, dbName)
+
+	// After restore is complete, the database users and passwords are reverted to those from the backup.
+	// This causes a mismatch between the actual database credentials and the Kubernetes User Secret.
+	// The PG operator does not automatically detect this drift, so it will not update the database passwords.
+	// To resolve this, we clear the User Secret, prompting the PG operator to regenerate new credentials
+	// and synchronize with the DB.
+	if err := resetUserSecrets(ctx, k8sClient, dbName, namespace); err != nil {
+		return fmt.Errorf("failed to reset user secrets: %w", err)
+	}
```
mayankshah1607, line 148: Just a heads up that with this workaround, we no longer require users to provide a password at the time of DB creation

internal/data-importer/cmd/pg/import.go
```diff
@@ -0,0 +1,429 @@
...
+	dbName, namespace string,
+) (string, error) {
+	pg := &pgv2.PerconaPGCluster{}
+	if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: dbName}, pg); err != nil {
+		return "", fmt.Errorf("failed to get PerconaPGCluster %s/%s: %w", namespace, dbName, err)
+	}
+	repoIdx := len(pg.Spec.Backups.PGBackRest.Repos) + 1
+	if repoIdx < 0 {
+		repoIdx = 0
+	}
```
recharte, line 177: We need some protection in case all 4 repos are already taken.

internal/data-importer/cmd/pg/import.go
```diff
@@ -0,0 +1,429 @@
...
+
+func preparePGBackrestSecret(
+	ctx context.Context,
+	c client.Client,
+	repoName string,
+	secretName string,
+	accessKeyId, secretAccessKey string,
+	dbName, namespace string,
+) error {
+	content := fmt.Sprintf(pgBackrestSecretContentTpl, repoName, accessKeyId, repoName, secretAccessKey)
```
recharte, line 235: Will this work if there are backup schedules in-place? Aren't we going to clear the secrets for those storages?

internal/data-importer/cmd/pg/import.go
```diff
@@ -0,0 +1,429 @@
...
+// This causes a mismatch between the actual database credentials and the Kubernetes User Secret.
+// The PG operator does not automatically detect this drift, so it will not update the database passwords.
+// To resolve this, we clear the User Secret, prompting the PG operator to regenerate new credentials
+// and synchronize with the DB.
+func resetUserSecrets(
+	ctx context.Context,
+	c client.Client,
+	dbName, namespace string,
+) error {
+	// We wrap this logic in a retry loop to reduce changes of conflicts or errors.
```
recharte, line 387:
```suggestion
	// We wrap this logic in a retry loop to reduce chances of conflicts or errors.
```

internal/data-importer/cmd/pg/import_test.go
```diff
@@ -0,0 +1,202 @@
...
+								{Name: "repo1"},
+								{Name: "repo2"},
+							},
+						},
+					},
+				},
+			},
+			want:    "repo3",
+			wantErr: false,
+		},
```
recharte, line 123: Let's add a case where all 4 repos are taken.

recharte (changes requested):

internal/data-importer/cmd/pg/import.go
mayankshah1607: Ah, you're right. I assumed that if it creates a completely new Secret and specifies it under the backups configuration, PG operator will somehow merge configs from both the Secrets (one provided by Everest operator and the other by the Data Importer). But that didn't happen, it used only the one provided here and complained that it needs the keys for the scheduled backups too.

I've fixed it now -- it updates the same Secret to include the S3 keys from the data importer

internal/data-importer/cmd/pg/import_test.go
mayankshah1607: Updated

PR #790, opened 2025-06-26
EVEREST-2006 | DataImporter improvements

mayankshah1607:
### Problem

1. Upstream restore objects created by data importers are orphan and cannot be deleted
2. DBR created for PGRestore created by data importers

### Solution
1. Attach owner ref to DB, so that it is automatically deleted when DB is deleted.
2. Add an annotation to skip creating DBR

### DIFF
--- a/internal/controller/databaseclusterrestore_controller.go
+++ b/internal/controller/databaseclusterrestore_controller.go
@@ -805,6 +805,11 @@ func (r *DatabaseClusterRestoreReconciler) tryCreatePG(ctx context.Context, obj 
 		return err
 	}
 
+	if val, ok := pgRestore.GetAnnotations()[consts.ManagedByDataImportAnnotation]; ok && val == consts.ManagedByDataImportAnnotationValueTrue {
+		// part of data import, do not create a DBR.
+		return nil
+	}
+
 	pg := &pgv2.PerconaPGCluster{}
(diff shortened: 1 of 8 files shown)


internal/controller/databaseclusterrestore_controller.go
```diff
@@ -805,6 +805,11 @@ func (r *DatabaseClusterRestoreReconciler) tryCreatePG(ctx context.Context, obj
 		return err
 	}
 
+	if val, ok := pgRestore.GetAnnotations()[consts.ManagedByDataImportAnnotation]; ok && val == consts.ManagedByDataImportAnnotationValueTrue {
+		// part of data import, do not create a DBR.
+		return nil
+	}
+
```
oksana-grishchenko, line 812: I don't get it why it is a problem that the controller creates a dbr when the upstream CR appears? Am I assuming that correctly that because of that confusing `Backup Source` field on the restores UI it was decided to get rid of the dbr CR and only keep the upstream CR?

oksana-grishchenko: Clarified, resolving the comment. As discussed on the DS, we should skip creating for the importer restore because it's getting created twice - once by importer, second - as a response to the new upstream restore appearance.

oksana-grishchenko (approved):

PR #809, opened 2025-07-08
EVEREST-2124 | Add PXC data importer e2e test

mayankshah1607:


tests/e2e/data-importers/pxc/11-assert.yaml
```diff
@@ -4,7 +4,7 @@ timeout: 300
 collectors:
   - command: kubectl get db/test-pxc-cluster -n ${NAMESPACE} -o yaml
   - command: kubectl get pxc/test-pxc-cluster -n ${NAMESPACE} -o yaml
-  - command: kubectl describe pods -n ${NAMESPACE}
```
mayankshah1607: What's the reason for removing this? The `describe` contains lots of useful information for why the pods do not come up.

maxkondr: the result of moving file 12 -> 11

PR #815, opened 2025-07-10
EVEREST-2153 | Graceful shutdown for Data importers

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-2153

Deleting a cluster in the middle of a DataImportJob results in dangling resources created by the Job.

**Cause:**

* These resources are cleaned-up only if the Job is allowed to succeeded.
* If deleted in the middle of the process, they are not able to be cleaned up correctly.

**Solution:**
* The DataImportJob process shall wait for SIGINT and SIGTERM signals to allow graceful shutdown.
* Specify a `terminationGracePeriod` in the Job pod to allow the Job pod to clean-up correctly.
* The clean-up routines shall use a different context from the parent since it may be cancelled
* Performing the cleanup also requires RBAC, so this PR refactors the controller such that the resources are cleaned up in specific order - first the Job is allowed to shut down, and only then the RBAC is resources are deleted.

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**


internal/consts/finalizers.go
```diff
@@ -33,7 +33,13 @@ const (
 	// DataImportJobRBACCleanupFinalizer is a finalizer that is used to clean up cluster-wide RBAC resources.
 	// This is needed because DataImportJob cannot own cluster-wide resources like ClusterRole and ClusterRoleBinding,
 	// as Kubernetes does not allow cluster-scoped resources to be owned by namespace-scoped resources.
+	//
+	// Deprecated: Use ordered-cleanup finalizer instead. This finalizer is kept for backward compatibility.
```
mayankshah1607, line 37: I've kept this finalizer assuming we will deliver this fix in 1.9.0. If we agree to deliver it in 1.8.0, then we can remove this finalizer completely

maxkondr (approved):

PR #821, opened 2025-07-14
Migrate tests from KIND to K3D Kubernetes cluster

maxkondr:
**CHANGE DESCRIPTION**
---
For some reason HAProxy unable to connect to DB nodes in case KIND cluster is used as K8S. Due to this problem PXC e2e test fails.

**Solution:**
Switch from using KIND cluster to K3D.

Addition:
- there are additional targets in Makefile: `k3d-cluster-up` to create K8S cluster and `k3d-cluster-down` to destroy K8S cluster
- by default K8S v1.33.2 is used in tests


tests/e2e/data-importers/pg/11-assert.yaml
```diff
@@ -1,9 +1,10 @@
 apiVersion: kuttl.dev/v1
 kind: TestAssert
-timeout: 300
+timeout: 600
 collectors:
   - command: kubectl get db/test-pg-cluster -n ${NAMESPACE} -o yaml
   - command: kubectl get perconapgcluster/test-pg-cluster -n ${NAMESPACE} -o yaml
+  - command: kubectl describe perconapgcluster/test-pg-cluster -n ${NAMESPACE}
```
oksana-grishchenko, line 7: do we need to have both `get perconapgcluster/test-pg-cluster` and `describe perconapgcluster/test-pg-cluster`?

.github/workflows/ci.yml
```diff
@@ -73,25 +91,35 @@ jobs:
...
-    env:
-      TOOLS_PATH: "/opt/tools/bin"
     steps:
-      - name: Install kubecltl
+      - name: Install TOOLS_PATH
         run: |
           mkdir -p $TOOLS_PATH
           echo $TOOLS_PATH >> $GITHUB_PATH
+
+      - name: Install kubecltl && kuttl
```
mayankshah1607, line 100:
```suggestion
      - name: Install kubectl && kuttl
```

mayankshah1607 (commented):
I think there's some clean-up from debugging, but otherwise LGTM

mayankshah1607:
PSMDB test seems to fail now, looks like we are running out of disk space:
```
message: '0/4 nodes are available: 1 node(s) didn''t match PersistentVolume''s node affinity, 3 node(s) had untolerated taint {node.kubernetes.io/disk-pressure: }. preemption: 0/4 nodes are available: 4 Preemption is not helpful for scheduling.'
```

I think the images from the PG run are being cached on the nodes, so by the time we get to the PSMDB test, we are left with little or no space.

One idea I have is that at the end of each test, we can add a step to clean-up unused images from all nodes like so:

```yaml
apiVersion: kuttl.dev/v1
kind: TestStep
timeout: 60
commands:
  - script: |
      NODES=$(kubectl get nodes -o jsonpath='{.items[*].metadata.name}')
      for NODE in $NODES; do
        docker exec $NODE ctr images prune -a
      done
```

mayankshah1607 (approved):

PR #852, opened 2025-08-07
EVEREST-107 Put back monitoring tests, use DEBUG flag

oksana-grishchenko:
**EVEREST-107 Put back monitoring integration tests**
---
**Problem:**
EVEREST-107

Use DEPLOY_TYPE env variable to skip PMM access validation

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**


maxkondr (commented):
I would suggest change ENV var name, because DEBUG is a general and widely used and has meaning "enable extra logging". but in our case we need to change the system behaviour

PR #860, opened 2025-08-19
EVEREST-2202 | Use shorter names for restore objects during data import

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-2202

Creating a import with a DB name longer than 16 characters results in the import failing.

**Cause:**

A long DB name results in a long name for the underlying restore objects, which in turn results in failed jobs. This is due to the fact that the name of the big restore object needs to be added as a label on the Job.

Take for example the DB name `mysql-importwithstream` which is 22 character long:
- this creates a PXCRestore object named `data-import-mysql-importwithstream-mysql`
- the PXC operator then creates a Job named `restore-job-data-import-mysql-importwithstream-mysql-importwithstream` which exceeds 63 characters
- this long string needs to be then attached as a label in the Job, but kubernetes does not allow you to have strings longer than 63 character as the label key.

**Solution:**

Since we cannot control how the operators manage labels and names, we shall user a shorter name for the PXC and PSMDB restore objects. Earlier we were using a name of the format `data-import-[DB NAME]`. Now we will calculate an md5 hash out of this string, and use the first 8 characters.

Note:
* the name of the PXC and PSMDB restore objects created during imports are of no use, so it should be okay to use such identifiers as names. The parent `DataImportJob` still uses a human friendly name, which is what is ultimately displayed to users
* We shall attach label to the restore objects to identify which DB they belong to. While this doesn't have any immediate use-case, it can be helpful in the future for debugging
* Nothing to change for PG since we do not use PGRestore for data import

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/internal/data-importer/cmd/pxc/import.go
+++ b/internal/data-importer/cmd/pxc/import.go
@@ -18,8 +18,6 @@ package pxc
 
 import (
 	"context"
-	"crypto/md5"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"os"
@@ -41,6 +39,7 @@ import (
 	everestv1alpha1 "github.com/percona/everest-operator/api/v1alpha1"
 	"github.com/percona/everest-operator/api/v1alpha1/dataimporterspec"
 	"github.com/percona/everest-operator/internal/consts"
+	"github.com/percona/everest-operator/internal/data-importer/utils"
 )
 
 // Cmd is the command for running PXC import.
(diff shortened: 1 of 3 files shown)


internal/data-importer/cmd/pxc/import.go
```diff
@@ -110,7 +112,11 @@ func runPXCImport(ctx context.Context, configPath string) error {
 		}
 	}()
 
-	pxcRestoreName := "data-import-" + dbName
+	// Create an 8-character long hash for the restore name.
+	// This is done to ensure that long DB names are shortened to not
+	// exceed the 63-character limit for Kubernetes labels.
+	nameHash := md5.Sum([]byte("data-import-" + dbName))
```
oksana-grishchenko, line 118: I would suggest to move it to some common function in order to avoid duplication. We don't have a good ready-to-use file to place it into though... we could have something for common funcitons under `internal/data-importer/cmd`. Up to you, not so critical

oksana-grishchenko (approved):

PR #861, opened 2025-08-21
EVEREST-2177 Add initContainer

oksana-grishchenko:
**Add the migrator initContainer**
---
**Problem:**
EVEREST-2177

In this PR:

1. **Added a new binary for migrator** 
Migrator is included as a separate binary with its own endpoint into the everest-operator image. Everest-operator now has initContainer that always tries to run the migrator. The migrator uses the Lease object to track wether the migration was already done (by checking annotations which contain the last version) and if so it does nothing. If the migration wasn't done, the migrator checks if the k8s type is EKS and if so - adds the loadBalancerConfigName = "eks-default" to all exposed DB clusters. 

3. **Removed the "Unstructured" layer.**  
Previously when Everest was checking the k8s type it was getting the `StorageClassList` using `Unstructured`. It was so seems the very beginning. After a research it seems to me that there is no actual need to use it, we can get the `StorageClassList` directly. My assumption that it was so to support the very old k8s types, which we don't support in Everest anyway yet. I removed this `Unstructured` layer because it complicates unit-testing a lot.

**Related pull requests**

- https://github.com/percona/everest/pull/1590
- https://github.com/percona/percona-helm-charts/pull/617

**Cause:**

**Solution:**

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@
 FROM golang:1.24 AS builder
 ARG TARGETOS
 ARG TARGETARCH
+ARG FLAGS
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
@@ -28,7 +29,7 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o ma
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o data-importer internal/data-importer/main.go
 
 # Build migration binary
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o migrator internal/migrate/main.go
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -ldflags "${FLAGS}" -o migrator internal/migrate/main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,12 +27,16 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o ma
 # Build data importer CLI
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o data-importer internal/data-importer/main.go
 
+# Build migration binary
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o migrator internal/migrate/main.go
+
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
 COPY --from=builder /workspace/data-importer .
+COPY --from=builder /workspace/migrator .
 USER 65532:65532
 
-ENTRYPOINT ["/manager"]
+ENTRYPOINT ["/manager", "/migrator"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,4 +39,4 @@ COPY --from=builder /workspace/data-importer .
 COPY --from=builder /workspace/migrator .
 USER 65532:65532
 
-ENTRYPOINT ["/manager", "/migrator"]
+ENTRYPOINT ["/manager"]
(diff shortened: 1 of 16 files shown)


internal/controller/common/helper.go
```diff
@@ -121,28 +121,20 @@ func GetOperatorVersion(
...
-	})
 	storageList := &storagev1.StorageClassList{}
 
-	err := c.List(ctx, unstructuredResource)
-	if err != nil {
-		return clusterType, err
-	}
-	err = runtime.DefaultUnstructuredConverter.
-		FromUnstructured(unstructuredResource.Object, storageList)
+	err := c.List(ctx, storageList)
```
oksana-grishchenko, line 126: The Unstructured layer seems to be a legacy that complicates unit testing. Please let me know if I miss something and we actually need it

Dockerfile
```diff
@@ -27,12 +27,16 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o ma
...
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
 COPY --from=builder /workspace/data-importer .
+COPY --from=builder /workspace/migrator .
 USER 65532:65532
 
-ENTRYPOINT ["/manager"]
+ENTRYPOINT ["/manager", "/migrator"]
```
mayankshah1607, line 42: Is this a valid command for the ENTRYPOINT? How would this work? Also do we actually need it, since the init container will explicitly specify the command to run

(For context, we don't have the `data-importer` command in the ENTRYPOINT either but it still works)

internal/migrate/migrator/migrator.go
```diff
@@ -0,0 +1,186 @@
...
+		if rerr != nil {
+			return
+		}
+
+		lease.Annotations[lastMigrationVersionAnnotationName] = m.currentVersion
+
+		err = m.client.Update(ctx, lease)
+		if err != nil {
+			rerr = fmt.Errorf("unable to update lease: %w", err)
+		}
```
recharte, line 91: I'm hesitant about running this logic inside a defer block. If the lease update fails, the init container will exit successfully as long as the migration itself worked. This means the Everest operator will continue to run, and on the next pod restart, the migration will execute again because the lease wasn't updated.

The core question is whether we want the operator to run if the migration succeeded but the lease update failed, or if we should force it to wait for a successful lease update before proceeding.

In both options, we need to make sure that the migration logic is idempotent because a lease update failure would lead to the migration running again eventually.

internal/migrate/migrator/migrator.go
```diff
@@ -0,0 +1,186 @@
...
+	return lease, nil
+}
+
+// This is the list of actions to be performed in the operator init container during the migration to the new
+// Everest version. Update this method for each next release
+//
+// For 1.9.0 Everest needs to patch all existing DBs if they are exposed and running on EKS
+// to assign the default .spec.proxy.expose.loadBalancerConfigName value.
+// Returns error and the additional info if needed.
+func (m *Migrator) migrateResources(ctx context.Context) (string, error) {
```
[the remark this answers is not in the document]
oksana-grishchenko: I thought about it but then the FBs are not going to work. From the other side, the FB versions start with `1.10000.0` and we could always allow migrations for such versions. WDYT?
It could be confusing after we release `1.9.0` that in the FB we will still have the migration, but we should cleanup the migration logic anyway after we release `1.9.0`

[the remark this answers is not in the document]
recharte: https://perconadev.atlassian.net/browse/EVEREST-2225

internal/migrate/migrator/migrator.go
oksana-grishchenko: > If the lease update fails, the init container will exit successfully as long as the migration itself worked. 

I don't think it's true. If the Lease update fails then the whole `Migrate()` function will fail (because in the `defer` we redefine the `rerr` which is being returned from the function). And then since the `Migrate()` function returns an error, `exit(1)` is going to be called and the initContainer will fail. Added a unit-test illustrating the failure

recharte: You are right, I missed the rerr definition in the function signature.

config/manager/manager.yaml
```diff
@@ -57,6 +57,19 @@ spec:
         # versions < 1.19 or on vendors versions which do NOT support this field by default (i.e. Openshift < 4.11 ).
         # seccompProfile:
         #   type: RuntimeDefault
+      initContainers:
+        - name: crs-migration
+          env:
+            - name: EVEREST_VERSION
+              value: latest
```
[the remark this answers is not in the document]
oksana-grishchenko: UPD: Resolved by using build flags as we do for API server

recharte (approved):

mayankshah1607 (approved):

PR #868, opened 2025-08-28
EVEREST-2233 Set ImagePullPolicy=IfNotPresent for upstream DB images

maxkondr:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-2233

Upstream DB CRs in most cases have `imagePullPolicy=Always` even for DB engine images. That leads to the following problems:

1. DB cluster restoration after some node failure takes more time because the images need to be re-fetched on each Pod creation.
2. Big companies/users that have a common external IP may face with "DockerHub Rate Limits" issues.
2.1. Image download speed  from DockerHub is not stable and may vary in time -> increases DB cluster restoration time after Node failure.
4. Our tests that create/destroy DB clusters massively are unstable due to unpredictable Pod creation time and "Rate limit" problems

**Solution:**
The images are tagged and  stable so there is no need to re-fetch them each time (nothing new will appear in them). So it is safe to set `imagePullPolicy=IfNotPresent`. 
That allows to decrease DB cluster restoration time and decreasing the chance to hit "DockerHub Rate Limit". In addition in our tests we can pre-upload images into K8S to stabilise Pods creations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,14 +189,17 @@ test-e2e-data-importer-pxc: docker-build k3d-upload-image ## Run e2e/data-import
 	. ./tests/vars.sh && kubectl kuttl test --config ./tests/e2e/kuttl-data-importer.yaml --test pxc
 
 .PHONY: k3d-cluster-up
-k3d-cluster-up: ## Create a K8S cluster for testing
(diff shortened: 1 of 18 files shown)


recharte (approved):
:pray:

PR #869, opened 2025-08-29
EVEREST-107 Disable wsl_v5 linter

oksana-grishchenko:
**Disable wsl_v5 linter**
---
**Problem:**
EVEREST-107

As agreed between the BE devs, removing the linter. It is too annoying suggesting to add empty lines to the files you touched. It does not bring that much value but it takes time to fix the broken CI.


mayankshah1607 (approved):
thank you!

recharte (approved):

PR #870, opened 2025-08-29
EVEREST-2264 Handle annotations removal

oksana-grishchenko:
**Handle annotations removal**
---
**Problem:**
EVEREST-2264

Everest wasn't handling properly the case of switching from `LoadBalancer` expose type to `ClusterIP`. 
However as it turned out, the annotations can be effectively removed only for `pg`, but the `pxc` and `psmdb` operators do not allow to delete the annotations by design (there was some reason for it which the Cloud team can't remember so far). 

As discussed with the Cloud team, there should be no harm in keeping the annotations when we change to the expose type to ClusterIP, they just will be ignored. So looks like we need to accept it as a limitation in Everest until it's fixed in the upstream. 

psmdb ticket https://perconadev.atlassian.net/browse/K8SPSMDB-1441
pxc ticket https://perconadev.atlassian.net/browse/K8SPXC-1709 

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/internal/controller/providers/psmdb/applier.go
+++ b/internal/controller/providers/psmdb/applier.go
@@ -262,6 +262,7 @@ func (p *applier) exposeShardedCluster(expose *psmdbv1.MongosExpose) error {
 	switch database.Spec.Proxy.Expose.Type {
 	case everestv1alpha1.ExposeTypeInternal:
 		expose.ExposeType = corev1.ServiceTypeClusterIP
+		expose.ServiceAnnotations = map[string]string{}
 	case everestv1alpha1.ExposeTypeExternal:
 		expose.ExposeType = corev1.ServiceTypeLoadBalancer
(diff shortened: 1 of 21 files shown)


internal/controller/providers/psmdb/applier.go
```diff
@@ -282,8 +283,9 @@ func (p *applier) exposeDefaultReplSet(expose *psmdbv1.ExposeTogglable) error {
 	database := p.DB
 	switch database.Spec.Proxy.Expose.Type {
 	case everestv1alpha1.ExposeTypeInternal:
-		expose.Enabled = false
+		expose.Enabled = true
```
recharte, line 286: Is this a mistake?

oksana-grishchenko: no, it is intended. the expose should be enabled, it's just the type has changed. I noticed that when we switch from LoadBalancer to ClusterIP, the actual service type in k8s was never changed, and this was what fixed it, now the type is changed as expected.

PR #871, opened 2025-09-03
EVEREST-2268 Handle existing providers annotations

oksana-grishchenko:
**Handle existing providers annotations**
---
**Problem:**
EVEREST-2268

As it turned out during discussion with the cloud team there are 3 issues:
1. annotations can't be completely deleted from `pxc` and `psmdb` services. Assigning an empty map to the expose.annotations section is not efficient by design. 
2. the cloud providers could set their own annotations to LoadBalancers independently and we need to keep them
3. if we use any custom annotations in `pxc` and `psmdb` they will rewrite the existing annotations in the service. To avoid that we need to put the annotations we want to persist to the `ignore` section of the cluster spec. 

**Solution:**
- To allow annotations deletion, from now on the everest-operator always applies the `everest.percona.com/custom-annotations=true` annotation to the exposure section of the upstream PR. It does nothing but takes place in the annotations map so that we could cleanup the custom annotations list efficiently, overcoming the upstream limitation. 
- everest-operator figures out the list of the annotations already applied to the exposure service and puts all annotations that are not listed in the `expose.annotations` of the upstream CR to the `ignore` section of the upstream CR . Only then we update the upstream CR `expose.annotation` section itself. So we make sure that we populated the `ignore` section before we update the `expose.annotations` section to avoid overwriting the existing annotations in the services. 
- DB controller watches Services annotations update and reconciles the related DB in case the service annotations have changed

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/internal/controller/common/exposure_annotations.go
+++ b/internal/controller/common/exposure_annotations.go
@@ -1,0 +1,120 @@
+package common
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	everestv1alpha1 "github.com/percona/everest-operator/api/v1alpha1"
+	"github.com/percona/everest-operator/internal/consts"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var errNoSvcFound = errors.New("no expose service found")
+
+// ReconcileExposureAnnotations returns the annotations to apply to the upstream cluster, the list of annotations keys to ignore to avoid overriding and an error
+func ReconcileExposureAnnotations(ctx context.Context, c client.Client, db *everestv1alpha1.DatabaseCluster, upstreamAnnotations map[string]string, componentName string) (desiredUpstreamAnnotations map[string]string, ignore []string, err error) {
+	serviceAnnotations, err := getExposeServiceAnnotations(ctx, c, db, componentName)
+
+	if err != nil {
+		// if there is no service available yet, we shouldn't populate the upstream annotations & ignore
+		if errors.Is(err, errNoSvcFound) {
+			return map[string]string{}, []string{}, nil
+		}
+		return nil, nil, err
+	}
+	everestAnnotations, err := getAnnotations(ctx, c, db)
+	if err != nil {
+		return nil, nil, err
+	}
+	desiredUpstreamAnnotations = addDefaultAnnotation(everestAnnotations)
+
+	return desiredUpstreamAnnotations, getUniqueKeys(serviceAnnotations, upstreamAnnotations), nil
+}
+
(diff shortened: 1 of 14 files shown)


internal/controller/common/exposure_annotations.go
```diff
@@ -0,0 +1,143 @@
...
+		annotations = make(map[string]string)
+	}
+	if _, ok := annotations[consts.DefaultEverestExposeServiceAnnotationsKey]; !ok {
+		annotations[consts.DefaultEverestExposeServiceAnnotationsKey] = consts.DefaultEverestExposeServiceAnnotationsValue
+	}
+	return annotations
+}
+
+// returns the keys that appears only in minuend.
+func getUniqueKeys(minuend, subtrahend map[string]string) []string {
```
recharte, line 129: Today I learned what minuend and subtrahend mean :see_no_evil: thanks Oksana :joy:

recharte (approved):
I think the end solution ended up being quite elegant, well done Oksana :clap:

oksana-grishchenko:
The lbc integration tests are failing bc now when the logic has changed, we need the LB service to be in k8s before we update the upstream annotations. However since the tests do not block testing the main logic, it was decided to merge the PR to the FB and move it to QA and then fix the failing tests in the Feature branch with no rush.

internal/controller/common/exposure_annotations.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, nil, err
+	}
+	everestAnnotations, err := getAnnotations(ctx, c, db)
+	if err != nil {
+		return nil, nil, err
+	}
+	desiredUpstreamAnnotations = addDefaultAnnotation(everestAnnotations)
+
+	return desiredUpstreamAnnotations, getUniqueKeys(serviceAnnotations, upstreamAnnotations), nil
+}
```
mayankshah1607, line 59: It was already discussed with Diogo on Slack, but I'm leaving a comment here for everyone's awareness:

There is a bug here in this code that causes keys to not be removed correctly on the Service when removing it from the LBC. When you remove a key from the LBC, it somehow lands up on the `ignoreAnnotations` list. The issue is that when a key is removed from the LBC, it gets removed from the upstream CR annotations as expected. But now we rely on the upstream operator to reconcile the removal onto the Service. However, if Everest performs another reconciliation before the upstream operator, it sees that the removed key is still present on the Service, but not on the CR annotations. Based on the current logic, it then decides to put this removed key onto the `ignoreAnnotations`, and the key is never removed.

The problem in this function is that it wrongly assumes that the state of the Service it observes is always up-to-date, when in fact it is possible that this may not be the case for a couple of reasons - reconciliation delays, reconciliation errors, or even if the upstream operator has crashed.

PR #891, opened 2025-09-29
EVEREST-2023 | Fix databasecluster controller reconciliation logic

mayankshah1607:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-2023

Inconsistent reconciliation logic in database cluster controller.

**Cause:**

In the current implementation, the DBC controller uses `controllerutil.CreateOrUpdate` for reconciliation. This helper first fetches and populates the DB object with the live state from k8s, then runs the mutate function, and finally performs an Update if changes are detected.

This effectively resets the default spec in set by the provider to the live object's spec before applying any mutations. But our provider code assumes each reconciliation constructs a fresh desired spec and compares it to the live object, which conflicts with CreateOrUpdate’s behaviour. As a result, fields computed in the provider can be wrongly overwritten by the fetched object, leading to unexpected drifts.

**Solution:**
- Implement a `ResetDefaults` method in the applier. This will be called first during in the mutation function, and will reset the spec of the upstream DB>
- Additionally, in each mutate step (Engine(), Proxy(), etc.) we shall first reset the corresponding part of the spec before applying changes. This way we ensure that the reconciliation process always builds a fresh spec based on the `DatabaseCluster` and ignoring anything that was previously set on the upstream object.

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/internal/controller/providers/pg/applier.go
+++ b/internal/controller/providers/pg/applier.go
@@ -142,6 +142,9 @@ func (p *applier) Engine() error {
 		return errors.Join(err, errors.New("could not update PG config"))
 	}
 
+	// Even though we call ResetDefault() as the first step in the mutation process,
+	// we still reset the spec here to protect against the defaults being unintentionally changed
+	// from a previous mutation.
 	pg.Spec.InstanceSets = defaultSpec().InstanceSets
 
 	pg.Spec.InstanceSets[0].Replicas = &database.Spec.Engine.Replicas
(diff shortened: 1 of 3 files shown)


internal/controller/providers/pg/applier.go
```diff
@@ -136,6 +141,9 @@ func (p *applier) Engine() error {
 	if err := p.updatePGConfig(pg, database); err != nil {
 		return errors.Join(err, errors.New("could not update PG config"))
 	}
+
+	pg.Spec.InstanceSets = defaultSpec().InstanceSets
+
```
maxkondr, line 146: why is it required? We already set the default spec with InstanceSets to `p.PerconaPGCluster.Spec`

internal/controller/providers/pg/applier.go
```diff
@@ -180,6 +188,8 @@ func (p *applier) Proxy() error {
 	pg := p.PerconaPGCluster
 	database := p.DB
 
+	pg.Spec.Proxy = defaultSpec().Proxy
```
maxkondr, line 191: the same

internal/controller/providers/pg/applier.go
mayankshah1607: Where? Do you mean in `ResetDefaults()`?

PR #913, opened 2025-12-10
EVEREST-2359 pg 2.8.0

oksana-grishchenko:
**PG 2.8.0 support**
---
**Problem:**
EVEREST-2359

Note: it was decided not to integrate with PG `2.8.1` in this release since this patch release bumped all the supported engine version which makes it unable to upgrade from `2.7.0` to `2.8.1` directly, so for this version we'll keep the `2.8.0` support only.  

**Changes:**
- removing PVC-based repo1 for clusters created with PG operator 2.8.0. For clusters, created by earlier versions, the pvc repo will be kept. DB restart is expected after upgrading the CR version since there are some changes applied between versions to the sts configuration. 

**Workarounds applied:**
1. **Restore to a new cluster with no repos.** If there is no repos defined, we should set `pg.spec.backups.enabled = false` which causes the cluster eventually gets stuck in the `Initialization` status after the successful restore. The `pg.status` shows 
```
Reconciliation is paused: please fill in spec.backups or add the [postgres-operator.crunchydata.com/authorizeBackupRemoval](http://postgres-operator.crunchydata.com/authorizeBackupRemoval) annotation to authorize backup removal.
```
in that case while there is no error when any repo is defined (e.g. when there is a schedule)

**The workaround:** keep a "dummy" repo during restoration. We already did it - we kept the storage used in the running restore. But now we keep it not just until the restore is complete but until the cluster has the `dataSource` field.

2. **Importing data to a cluster with no repos.** 
If there is no repos defined, we should set `pg.spec.backups.enabled = false` which causes the cluster gets stuck in the `Importing` state while the pg operator logs show:
```
Job.batch \"cluster2-pgbackrest-restore\" is invalid: spec.template.spec.initContainers[0].image: Required value
```
in that case while there is no error when any repo is defined (e.g. when there is a schedule)

**The workaround:** keep a "dummy" repo during restoration. When configuring import, use the same repo in the `pg.spec.dataSource` and in the `pg.spec.backups` sections.

**Related pull requests**

- https://github.com/percona/everest-catalog/pull/71
- https://github.com/percona/everest/pull/1721 

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/test/e2e/operator-upgrade/pg/10-create-cluster.yaml
+++ b/test/e2e/operator-upgrade/pg/10-create-cluster.yaml
@@ -15,5 +15,7 @@ spec:
       cpu: "1"
       memory: 1G
   proxy:
+    expose:
+      type: internal
     type: pgbouncer
     replicas: 3


internal/controller/everest/providers/pg/applier.go
```diff
@@ -1161,6 +1175,21 @@ func (p *applier) reconcilePGBackupsSpec() (pgv2.Backups, error) {
 	return newBackups, nil
 }
 
+// Keep the PVC storage only if it was defined before or if PG less than 2.7.0 is used.
```
recharte:
```suggestion
// Keep the PVC storage only if it was defined before or if PGO less than 2.8.0 is used.
```

internal/controller/everest/providers/pg/pg_test.go
recharte: This file is crazy 🤯 but kudos on how thorough you were  👏

recharte (approved):

PR #922, opened 2026-01-28
Support go-templating for dynamic value injection in lbc

Ankit152:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-1817

Support Go templating in LoadBalancerConfig annotations for dynamic value injection.
Fixes: https://github.com/openeverest/openeverest/issues/1817

**Related pull requests**

- N/A

**Cause:**

**Solution:**

The openeverest-operator currently uses a [helper function](https://github.com/openeverest/openeverest-operator/blob/741cc0387bdd3f3d46ad561a5a35af1dd68b4e60/internal/controller/everest/common/helper.go#L681-L696) to fetch annotations. This logic should be updated to:

1. Iterate over the annotations defined in the LBC.
2. Check if the value contains template syntax.
3. Execute the template passing the relevant `DatabaseCluster` struct.
4. Apply the resolved string as the final annotation value.

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**

### DIFF
--- a/internal/controller/everest/common/helper.go
+++ b/internal/controller/everest/common/helper.go
@@ -26,6 +26,7 @@ import (
 	"reflect"
 	"slices"
 	"strings"
+	"text/template"
 
 	"github.com/AlekSi/pointer"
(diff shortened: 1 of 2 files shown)


recharte (approved):
Changes look great 🚀  thanks

PR #940, opened 2026-03-20
Fixing DatabaseCluster proxysql resource requests

### LINKED ISSUE #938: panic: assignment to entry in nil map when creating DatabaseCluster with proxy.type=proxysql
When creating a `DatabaseCluster` with `proxy.type: proxysql`, the `everest-operator` panics with `assignment to entry in nil map` inside `applyProxySQLCfg`. The cluster gets stuck in `creating` status indefinitely and never provisions any pods.

This bug was introduced in **v1.3.0** (#552) and is present through the current **v1.14.0**.

## Version
- OpenEverest: v1.14.0
- Kubernetes: v1.34.3

## Steps to Reproduce
1. Install OpenEverest v1.14.0
2. Create a `DatabaseCluster` with `proxy.type: proxysql` and explicit proxy resources:

```yaml
apiVersion: everest.percona.com/v1alpha1
kind: DatabaseCluster
metadata:
  name: mysql-pxc
  namespace: everest-db
spec:
  engine:
    type: pxc
    replicas: 3
    version: "8.0.39-30.1"
    resources:
      cpu: "1"
      memory: 2G
    storage:
      class: standard
      size: 20Gi
    userSecretsName: everest-secrets-mysql-pxc
  proxy:
    type: proxysql
    replicas: 1
    expose:
      type: ClusterIP
    resources:
      cpu: 200m
      memory: 200M
```

3. Observe: `kubectl get databasecluster` shows status `creating` forever, no pods are created.

## Error

From `kubectl logs -n everest-system deployment/everest-operator`:

```
ERROR   Observed a panic   {"panic": "assignment to entry in nil map", ...
goroutine 791 [running]:
...
github.com/percona/everest-operator/internal/controller/everest/providers/pxc.(*applier).applyProxySQLCfg(0xc000c9f548)
	/workspace/internal/controller/everest/providers/pxc/applier.go:706 +0x8e8
github.com/percona/everest-operator/internal/controller/everest/providers/pxc.(*applier).Proxy(0xc000c9f548)
	/workspace/internal/controller/everest/providers/pxc/applier.go:320 +0x198
```

### Primary bug (introduced in v1.3.0, commit b731282 / EVEREST-1555)

In `defaultSpec()`, the `ProxySQL` spec initializes `Resources.Limits` but **never initializes `Resources.Requests`**, leaving it as a nil map:

```go
ProxySQL: &pxcv1.ProxySQLSpec{
    PodSpec: pxcv1.PodSpec{
        Resources: corev1.ResourceRequirements{
            Limits: corev1.ResourceList{...},  // initialized
            // Requests is nil, never initialized
        },
    },
},
```

Commit `b731282` (EVEREST-1555) added `shouldUpdateRequests` and started writing into `proxySQL.Resources.Requests[corev1.ResourceCPU]` inside `applyProxySQLCfg`, but never initialized the map in `defaultSpec()`. Any `DatabaseCluster` with non-zero `proxy.resources.cpu` or `proxy.resources.memory` hits the panic.

> **Note:** if `proxy.resources` is omitted entirely (zero values), the `if !p.DB.Spec.Proxy.Resources.CPU.IsZero()` guard prevents the panic, so clusters without explicit proxy resources may not be affected.

### Secondary bug (current main)

`applyProxySQLCfg` references **HAProxy's** resources in the condition that guards ProxySQL's `Requests` assignment, a copy-paste error:

```go
// Wrong: reads HAProxy resources instead of ProxySQL
if shouldUpdateRequests ||
    p.currentPerconaXtraDBClusterSpec.HAProxy.Resources.Requests.Cpu().
        Equal(p.DB.Spec.Proxy.Resources.CPU) {
    proxySQL.Resources.Requests[corev1.ResourceCPU] = ...
}
```

## Additional Error (after removing proxy.resources)

If `proxy.resources` is omitted (working around the nil map panic), a second error surfaces from the PXC operator:

```
Error: validate cr: ProxySQL: volumeSpec should be specified
```

The `everest-operator` does not populate `volumeSpec` for ProxySQL when translating the `DatabaseCluster` CRD into a `PerconaXtraDBCluster` resource. This means **ProxySQL is completely unusable** via the `DatabaseCluster` CRD in v1.14.0, both bugs must be fixed together.

There is no workaround for this second bug: any manual patch to the `PerconaXtraDBCluster` resource is reverted by the `everest-operator` on the next reconcile loop (~5s). The `DatabaseCluster` CRD does not expose a `volumeSpec` field for the proxy, so there is no supported way to pass it through.

## Workaround

Use `proxy.type: haproxy` instead

Ankit152:
**CHANGE DESCRIPTION**
---
**Problem:**
Issue: https://github.com/openeverest/openeverest-operator/issues/938

1. When creating a DatabaseCluster with `proxy.type: proxysql`, the everest-operator panics with assignment to entry in nil map inside `applyProxySQLCfg`. The cluster gets stuck in creating status indefinitely and never provisions any pods.
2. `applyProxySQLCfg` references HAProxy's resources in the condition that guards ProxySQL's Requests assignment, a copy-paste error.
3. If `proxy.resources` is omitted (working around the nil map panic), a second error surfaces from the PXC operator. The everest-operator does not populate volumeSpec for ProxySQL when translating the DatabaseCluster CRD into a PerconaXtraDBCluster resource.

**Cause:**
1. In `defaultSpec()`, the `ProxySQL` spec initializes `Resources.Limits` but never initializes `Resources.Requests`, leaving it as a nil map:
```
ProxySQL: &pxcv1.ProxySQLSpec{
    PodSpec: pxcv1.PodSpec{
        Resources: corev1.ResourceRequirements{
            Limits: corev1.ResourceList{...},  // initialized
            // Requests is nil, never initialized
        },
    },
},
```
2. Copy paste error:
```
// Wrong: reads HAProxy resources instead of ProxySQL
if shouldUpdateRequests ||
    p.currentPerconaXtraDBClusterSpec.HAProxy.Resources.Requests.Cpu().
        Equal(p.DB.Spec.Proxy.Resources.CPU) {
    proxySQL.Resources.Requests[corev1.ResourceCPU] = ...
}
```
3. The `DatabaseCluster` CRD does not expose a `volumeSpec` field for the `proxy`, so there is no supported way to pass it through.

**Solution:**

- Fix nil map initialisation error
- Update appropriate resource mapping for ProxySQL
- Add support for `Storage` in `DatabaseCluster` CRD to expose `volumeSpec` field for the `proxy`

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**


internal/controller/everest/providers/pxc/applier.go
```diff
@@ -135,6 +135,38 @@ func configureStorage(
...
 
+func configureProxySQLStorage(
+	ctx context.Context,
+	c client.Client,
+	desired *pxcv1.PerconaXtraDBClusterSpec,
+	db *everestv1alpha1.DatabaseCluster,
+) error {
+
+	getCurrentProxySQLStorageSize := func() resource.Quantity {
+		if db.Spec.Proxy.Storage == (everestv1alpha1.Storage{}) || db.Spec.Proxy.Storage.Size.IsZero() {
```
recharte: Why do we need the first condition here? If `db.Spec.Proxy.Storage` is an empty struct then `db.Spec.Proxy.Storage.Size.IsZero()` is true, right? Am I missing something?

internal/controller/everest/common/helper.go
```diff
@@ -1045,6 +1045,66 @@ func ConfigureStorage(
 	return nil
 }
 
+func ConfigureProxySQLStorage(
```
recharte: I don't think we should have this method. This is basically a duplicate of `ConfigureStorage()` with the only difference being that this one uses `db.Spec.Proxy.Storage` instead of `db.Spec.Engine.Storage`.

Maybe we could add one more parameter of type `everestv1alpha1.Storage` to the `ConfigureStorage()` signature for the desired storage. This way we can call the `ConfigureStorage()` with `db.Spec.Engine.Storage` for the existing use cases and with `db.Spec.Proxy.Storage` for this new use case.

What do you think?

This would also mean that we need to handle the case of an empty storage class in the caller but I think that's fine.

api/everest/v1alpha1/databasecluster_types.go
```diff
@@ -481,6 +483,10 @@ const (
 	ConditionTypeVolumeResizeFailed = "VolumeResizeFailed"
 	// ConditionTypeImportFailed is a condition type that indicates that the data import failed.
 	ConditionTypeImportFailed = "ImportFailed"
+	// ConditionTypeCannotResizeProxySQLVolume is a condition type that indicates that the ProxySQL volume cannot be resized.
+	ConditionTypeCannotResizeProxySQLVolume = "CannotResizeProxySQLVolume"
+	// ConditionTypeProxySQLVolumeResizeFailed is a condition type that indicates that the ProxySQL volume resize failed.
+	ConditionTypeProxySQLVolumeResizeFailed = "ProxySQLVolumeResizeFailed"
```
recharte: I don't think we need to have specific conditions for ProxySQL, we can just repurpose the existing `ConditionTypeCannotResizeVolume` and `ConditionTypeVolumeResizeFailed`.

recharte (changes requested):
Thank you @Ankit152, left a few suggestions.

internal/controller/everest/providers/pxc/applier.go
Ankit152: I am just trying to be safe here. Imagine a situation where a user intentionally or unintentionally didn't mention the size of the volume. In that case, we will still have a volume of 2Gi for ProxySQL.

api/everest/v1alpha1/databasecluster_types.go
Ankit152: Alright. It makes sense since we are reusing the `ConfigureStorage()` function. 👍🏼

internal/controller/everest/providers/pxc/applier.go
recharte: I agree with that protection.
My question was about the need for the first part of the condition, `db.Spec.Proxy.Storage == (everestv1alpha1.Storage{})`.
Wouldn't
```
if db.Spec.Proxy.Storage.Size.IsZero() {
```
cover everything?

I can't think of a case where `db.Spec.Proxy.Storage == (everestv1alpha1.Storage{})` is true and `db.Spec.Proxy.Storage.Size.IsZero()` is false. So I believe that the second condition suffices.
Am I missing something?

Ankit152: The reason why I put `db.Spec.Proxy.Storage == (everestv1alpha1.Storage{})` as condition is to remove any scope of panic occurring while trying to fetch a value of a non existing struct. 🙂

recharte: Now that you mention this I just figured out that you set the `Storage` as a mandatory param in [the spec](https://github.com/Ankit152/openeverest-operator/blob/658230c99822e4c583927193f238a208f21394c2/api/everest/v1alpha1/databasecluster_types.go#L292-L293). I think you should change it to be a pointer so that it can be nil in deployments that don't require storage for the proxy. Then it makes sense to have a `db.Spec.Proxy.Storage == nil` condition to protect against panics.

internal/controller/everest/common/helper.go
Ankit152: I have removed the `ConfigureProxySQLStorage` function and modified `ConfigureStorage` function according to the needs.

internal/controller/everest/providers/pxc/applier.go
Ankit152: Done.

api/everest/v1alpha1/databasecluster_types.go
```diff
@@ -289,6 +289,8 @@ type Proxy struct {
 	Replicas *int32 `json:"replicas,omitempty"`
 	// Config is the proxy configuration
 	Config string `json:"config,omitempty"`
+	// Storage is the proxy storage configuration
+	Storage *Storage `json:"storage"`
```
recharte: We need to explicitly make the field optional with a `// +optional` marker and `omitempty` struct tag otherwise we get errors like:
```
DatabaseCluster.everest.percona.com "mysql-32k" is invalid: [spec.proxy.storage: Invalid value: "null": spec.proxy.storage in body must be of type object: "null", \u003cnil\u003e: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
```

```suggestion
	// Storage is the proxy storage configuration
	// +optional
	Storage *Storage `json:"storage,omitempty"`
```

internal/controller/everest/providers/pxc/applier.go
```diff
@@ -692,6 +750,10 @@ func (p *applier) applyProxySQLCfg() error {
 	}
 	proxySQL.Image = image
 
+	if err := configureProxySQLStorage(p.ctx, p.C, &p.PerconaXtraDBCluster.Spec, &p.currentPerconaXtraDBClusterSpec, p.DB); err != nil {
```
recharte: Here we are passing `p.PerconaXtraDBCluster.Spec` to `configureProxySQLStorage` meaning that the ProxySQL storage spec is being applied directly to `p.PerconaXtraDBCluster.Spec`. However, on line 784 we are setting `p.PerconaXtraDBCluster.Spec.ProxySQL = proxySQL`, effectively overriding whatever `configureProxySQLStorage` did with the local `proxySQL` spec.
In order for `configureProxySQLStorage` to do its job, we should be editing `p.PerconaXtraDBCluster.Spec` in the entire scope of this `applyProxySQLCfg()` function, similarly to how `Engine()` works with its `configureStorage`.

recharte (changes requested):
@Ankit152 There are a few issues that are making this solution not work as intended, please see the comments below.
Also, please extend the [PXC core integration tests](https://github.com/openeverest/openeverest-operator/tree/main/test/integration/core/pxc) with one extra step (e.g. 60-create-cluster-proxysql.yaml) similar to the existing 10-create-cluster.yaml and 10-assert.yaml but adjusted for ProxySQL. This way we will be sure that the reconciliation is actually doing the mapping as intended. Also, don't forget to add another delete-cluster step to clean up (e.g. 70-delete-cluster.yaml).

api/everest/v1alpha1/databasecluster_types.go
recharte: This is what's causing the failures you see in the integration tests CI pipeline.

Ankit152: Yes. I also thought the same and did call out on yesterday's meeting. 
I was waiting for you to have a look before I make the necessary changes.

No issues. I will fix it. 🙂

test/integration/core/pxc/61-assert-proxysql.yaml
```diff
@@ -0,0 +1,160 @@
...
+metadata:
+  name: test-pxc-cluster
+spec:
+  backup:
+    pitr:
+      enabled: false
+      resources: {}
+      storageName: ""
+  proxysql:
+    enable: true
```
recharte: According to the [CI run](https://github.com/openeverest/openeverest-operator/actions/runs/24247553196/job/70804701831?pr=940) it seems like this is a typo and it should be `enabled`

Ankit152: Ahhh. My bad. Fixed.

internal/controller/everest/providers/pxc/applier.go
Ankit152: Fixed.

recharte (approved):
Hi @Ankit152 you did a great job 👏 
While testing I found a couple of things and thought it would be easier to push a fix than to request more changes. I hope you don't mind 🤗 

Here's a breakdown of the commits I just pushed:

- Regarding the kuttl tests, we should follow conventions of having one `xx-assert.yaml` file per each test step. Hence I:
  - Renamed `61-assert-proxysql.yaml` to `60-assert.yaml`. The sequence number is important for kuttl to correctly identify the test that corresponds to the failed assert step, in case of a failure. ae413aa1a9328da26f1284e3314744c75f402551
  - Added a missing assert for `70-delete-cluster.yaml`. 5ad34334ea9b02b76776cfe52a043701900e94c2
- We need to ensure that the `percona.com/delete-proxysql-pvc` finalizer is set. Otherwise, the PXCOperator won't clean up the PVCs and we'll have orphan PVCs after deletion. c2eb5cfaed0d52e60b7f9d737e30a852917407ad
- 755af7c48575f639e70ffbd0d5f9198c8e64be1a Completely unrelated with your PR, just a typo I came across.

Ankit152:
Hey @recharte 👋🏼 

It's totally fine. Thanks for highlighting and fixing the underlying issues, it's easier for me to follow-up. 😄

recharte:
@Ankit152 in case you are wondering what the next steps after doing a CRD change in the operator are, these 2 PRs will be very informative for you:
* https://github.com/openeverest/helm-charts/pull/26
  * Update the CRDs in the helm charts repo. This is a crucial step for the installation. This can be done by running `make crds-gen` (or the superset `make prepare-pr`)
* https://github.com/openeverest/openeverest/pull/2008
  * Update the API spec with the new CRD in `docs/spec/openapi.yaml`
  * Generate the go clients: `make generate`
  * Update the operator go mod so that the API can properly parse the new structs: `make update-dev-everest-operator`
  * Update the helm chart go mod so that dev builds of the everestctl install the helm chart that includes the recent CRD changes: `make update-dev-chart`

Ankit152:
Thank you for explaining and providing insights. I will remember this next time I make any CRD changes to the operator. 😉

PR #948, opened 2026-04-23
feat: add backup size to PSMDB

AdityaPimpalkar:
**CHANGE DESCRIPTION**
---
**Problem:**
[#1766](https://github.com/openeverest/openeverest/issues/1766)

**CHECKLIST**
---
**Helm chart**

**Jira**

**Tests**


internal/controller/everest/databaseclusterbackup_controller.go
```diff
@@ -395,6 +395,7 @@ func (r *DatabaseClusterBackupReconciler) getBackupStatus(
 		backupStatus.CompletedAt = psmdbCR.Status.CompletedAt
 		backupStatus.CreatedAt = &psmdbCR.CreationTimestamp
 		backupStatus.Destination = &psmdbCR.Status.Destination
+		backupStatus.Size = &psmdbCR.Status.Size
```
chilagrow, line 398: Let's add a test for this!

We use `kuttl` for testing Custom Resource changes, I think we could add an assertion that `.Size` field has some value.

The [backup_ondemand_psmdb/43-assert.yaml](https://github.com/openeverest/openeverest-operator/blob/main/test/integration/features/backup_ondemand_psmdb/43-assert.yaml) file asserts that required conditions are met upon successful backup by psmdb (Percona Server MongoDB). We can add an additional assertion for `.Size` field. 🤗

chilagrow (commented):
Thanks for your contribution! 😃 Let's add a test for the new feature

internal/controller/everest/databaseclusterbackup_controller.go
AdityaPimpalkar: @chilagrow Done! Thanks for the help :)

recharte (approved):

PR #952, opened 2026-04-30
feat: apply proxy config to mongos in sharded mode

areebahmeddd:
**CHANGE DESCRIPTION**
---
**Problem:**
openeverest/openeverest#2094

`DatabaseCluster.spec.proxy.config` was effectively a no-op for PSMDB, even with sharding enabled there was no way to pass custom config to mongos.

**Related pull requests**

- N/A

**Cause:**
While building the mongos spec in `Proxy()`, `spec.proxy.config` was never mapped to `sharding.mongos.configuration`, so it always stayed empty.

**Solution:**
Forward `spec.proxy.config` to `mongos.configuration` when sharding is enabled so it actually takes effect. When sharding is disabled, it is ignored since no mongos is created.

**AI Model Disclosure:**
Used VSCode Copilot (Claude Sonnet 4.6) to understand the issue context and help generate a fix. Changes were self-reviewed and verified via:

```bash
go build ./...
make test
make static-check
````

## **CHECKLIST**

**Helm chart**

* [ ] Is the [helm chart](https://github.com/percona/percona-helm-charts/tree/main/charts/everest) updated with the new changes? (if applicable)

**Jira**

* [ ] Is the Jira ticket created and referenced properly?

**Tests**

* [x] Is an Integration test/test case added for the new feature/change?
* [x] Are unit tests added where appropriate?

### CODE REVIEW

**Review by Ankit152 — CHANGES_REQUESTED** [review]
I don't think we would need so many integration tests just for a simple two line change.

We can just have one extra integration test under `test/integration/core/psmdb` that tests sharding of mongose config.

File: test/integration/core/psmdb-sharded/10-create-sharded-cluster.yaml [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+  engine:
+    crVersion: 1.21.2
+    replicas: 1
+    resources:
+      cpu: "1"
+      memory: 4G
+    storage:
+      size: 25Gi
+    type: psmdb
+  sharding:
```
chilagrow, line 19: Let's create `test/integration/core/psmdb-sharded/` directory for this test. `psmdb` directory has tests for replica-sets only.

Please check recommended `Integration tests` suggestion in the issue 🤗

areebahmeddd: sure, i'll add integration tests for create, set, clear and delete 🙂

test/integration/core/psmdb-sharded/10-create-sharded-cluster.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: kuttl.dev/v1
+kind: TestStep
+timeout: 10
+---
+apiVersion: everest.percona.com/v1alpha1
+kind: DatabaseCluster
+metadata:
+  name: test-psmdb-sharded-cluster
+spec:
```
chilagrow, line 9: Could you assert `spec.sharding.mongos` contains expected values please.

internal/controller/everest/providers/psmdb/applier_test.go
```diff
@@ -1133,3 +1136,90 @@ func TestGetPMMResources(t *testing.T) {
...
+		name            string
+		proxyConfig     string
+		shardingEnabled bool
+		wantConfig      psmdbv1.MongoConfiguration
+	}{
+		{
+			name:            "Sharded cluster with proxy.config",
+			proxyConfig:     "operationProfiling:\n  mode: slowOp\n",
+			shardingEnabled: true,
+			wantConfig:      psmdbv1.MongoConfiguration("operationProfiling:\n  mode: slowOp\n"),
```
chilagrow: Do you need this test? They duplicate kuttl tests.

test/integration/core/psmdb-sharded/20-set-proxy-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+    enabled: true
+    shards: 1
+    configServer:
+      replicas: 1
+  proxy:
+    config: |
+      operationProfiling:
+        mode: slowOp
+    expose:
+      type: internal
```
chilagrow: Current change looks very busy and difficult to identify what has changed, let's update the only fields that needs change. This case, you should only change spec.proxy.config.
```suggestion
  proxy:
    config: |
      operationProfiling:
        mode: slowOp
```

test/integration/core/psmdb-sharded/30-clear-proxy-config.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1
+kind: TestStep
+timeout: 10
+commands:
+  - command: kubectl patch dbc/test-psmdb-sharded-cluster -n $NAMESPACE --type=merge -p '{"spec":{"proxy":{"config":""}}}'
```
chilagrow: Could you use the same format as `20-set-proxy-config.yaml‎` for consistency, instead of using kubectl please? 🤗

areebahmeddd:
ping 🙂 @chilagrow, addressed your feedback. hope it alligns with your expectations

test/integration/core/psmdb-sharded/10-assert.yaml
```diff
@@ -0,0 +1,78 @@
...
+  - celExpr: "has(db.metadata.finalizers)"
+    message: "db doesn't have finalizers"
+
+  - celExpr: "'everest.percona.com/upstream-cluster-cleanup' in db.metadata.finalizers"
+    message: "'everest.percona.com/upstream-cluster-cleanup' is absent in db.metadata.finalizers"
+
+  - celExpr: "'everest.percona.com/dbb-cleanup' in db.metadata.finalizers"
+    message: "'everest.percona.com/dbb-cleanup' is absent in db.metadata.finalizers"
+
+  - celExpr: "psmdb.spec.sharding.enabled == true"
```
chilagrow: This condition is already being asserted in below L76. I don't think we need to duplicate it here. We use cel expression for things beyond simple yaml match, that's often array items. 🤗

test/integration/core/psmdb-sharded/10-assert.yaml
```diff
@@ -0,0 +1,78 @@
...
+    message: "psmdb.spec.sharding.enabled is not true"
+
+  - celExpr: "psmdb.spec.replsets.size() == 1"
+    message: "psmdb.spec.replsets.size() is not 1"
+
+  - celExpr: "psmdb.spec.sharding.mongos.size == 1"
```
chilagrow: Same here, it's already asserted in L78. Please check other places too!

test/integration/core/psmdb-sharded/10-assert.yaml
```diff
@@ -0,0 +1,78 @@
...
+    message: "psmdb.spec.sharding.mongos.size is not 1"
+
+  - celExpr: "!has(psmdb.spec.sharding.mongos.configuration) || psmdb.spec.sharding.mongos.configuration == ''"
```
chilagrow: Is there a default value, what was it? Could you check that the test in 20-set-proxy-config.yaml step sets different from default value? This ensures that value is altered. 🤗

test/integration/core/psmdb-sharded/30-assert.yaml
```diff
@@ -0,0 +1,23 @@
...
+    kind: DatabaseCluster
+    name: test-psmdb-sharded-cluster
+    ref: db
+  - apiVersion: psmdb.percona.com/v1
+    kind: PerconaServerMongoDB
+    name: test-psmdb-sharded-cluster
+    ref: psmdb
+assertAll:
+  - celExpr: "!has(psmdb.spec.sharding.mongos.configuration) || psmdb.spec.sharding.mongos.configuration == ''"
+    message: "psmdb.spec.sharding.mongos.configuration is not empty"
```
chilagrow, line 19: Instead of using cel expression here, let's use assertion such as below

```
apiVersion: psmdb.percona.com/v1
kind: PerconaServerMongoDB
metadata:
  name: test-psmdb-sharded-cluster
spec:
  sharding:
    mongos:
      configuration: ""
```

test/integration/core/psmdb-sharded/20-set-proxy-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+---
+apiVersion: everest.percona.com/v1alpha1
+kind: DatabaseCluster
+metadata:
+  name: test-psmdb-sharded-cluster
+spec:
+  proxy:
+    config: |
+      operationProfiling:
+        mode: slowOp
```
chilagrow, line 13: > Bonus: a non-sharded test that sets spec.proxy.config and asserts the
field is not propagated anywhere (sanity check that we don't
accidentally write it to the replset).

Would you like to add a test in `integration/core/psmdb/21-set-proxy.yaml` to ensure that spec.proxy.config is ignored?

chilagrow (commented):
Thanks for your work! I have a few comments around assertion and extending test coverage to psmdb replset

areebahmeddd:
hello chi! addressed your comments! apologies for the long pr triage. the test cases are very hard to get right haha 😅

test/integration/core/psmdb-sharded/10-assert.yaml
chilagrow: Could you use assertion instead of using cel expression here please, like other places please?

test/integration/core/psmdb-sharded/20-assert.yaml
```diff
@@ -0,0 +1,20 @@
...
+    tail: 100
+---
+apiVersion: psmdb.percona.com/v1
+kind: PerconaServerMongoDB
+metadata:
+  name: test-psmdb-sharded-cluster
+spec:
+  sharding:
+    mongos:
+      configuration: "operationProfiling:\n  mode: slowOp\n"
```
chilagrow: Let's use multi-line string consistently for ease of readability. In 20-set-proxy-config.yaml sets multi-line let's also assert using multi-line 🤗 .

```
    configuration: |
      operationProfiling:
        mode: slowOp
```

test/integration/core/psmdb/21-assert.yaml
```diff
@@ -0,0 +1,27 @@
...
+    selector: control-plane=controller-manager
+    tail: 100
+resourceRefs:
+  - apiVersion: psmdb.percona.com/v1
+    kind: PerconaServerMongoDB
+    name: test-psmdb-cluster
+    ref: psmdb
+assertAll:
+  - celExpr: "!has(psmdb.spec.replsets[0].configuration) || !psmdb.spec.replsets[0].configuration.contains('maxIncomingConnections')"
+    message: "psmdb.spec.replsets[0].configuration contains proxy.config value"
```
chilagrow: Could you tell me why we are checking replsets? Isn't it sufficient to check that `spec.sharing.mongos` or `spec.sharing.mongos.configuration` is unset? I think it's the former.

test/integration/core/psmdb-sharded/30-assert.yaml
```diff
@@ -0,0 +1,20 @@
...
+    tail: 100
+---
+apiVersion: psmdb.percona.com/v1
+kind: PerconaServerMongoDB
+metadata:
+  name: test-psmdb-sharded-cluster
+spec:
+  sharding:
+    mongos:
+      configuration: ""
```
chilagrow: This integration test is failing on CI. https://github.com/openeverest/openeverest-operator/actions/runs/25854390167/job/76106667286?pr=952

```
    case.go:398: resource PerconaServerMongoDB:kuttl-test-rapid-cub/test-psmdb-sharded-cluster: .spec.sharding.mongos.configuration: key is missing from map
```

chilagrow (commented):
Thanks a lot for your effort. Let's make sure that test passes!

test/integration/core/psmdb-sharded/30-assert.yaml
areebahmeddd: pushed a possible fix! i think ci failed because mongos.configuration gets omitted when it's empty 🤔

also, dco was failing on my latest commit so had to force push (even tho all my commits have werified tag, weird.)

chilagrow (approved):
Thank you for your contribution! Great work

PR #955, opened 2026-05-19
fix: stop misclassifying plain S3 credentials as base64 (#2245)

### LINKED ISSUE #2229: No checks done when providing S3 details.
### What is the issue?

When trying to import from external backup using the documented steps:

https://openeverest.io/documentation/1.15.2/backups_and_restore/dataimporters/pg_dataimporter.html#how-to-import-external-postgresql-backups-using-the-openeverest-ui 

Using incorrect/fake information leads to a seemingly successful environment creation. However, the database will be forever stuck at "importing" state:

Image

### Steps to Reproduce

1. use the following:

Image

2. and you can also use a fake path:

Image

3. You will see all green, as if things are ok:

Image

4. Move through the rest of the steps as you normally would, and click on create database at the end.

5. the database will be initializing, and then stuck at importing.

### Relevant Logs / Output

_No response_

### OpenEverest Version

OpenEverest v1.15.2

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

Client Version: v1.36.0
Kustomize Version: v5.8.1
Server Version: v1.35.3-gke.1389000

### Possible Solution

Proposed solution: actually test data that was input by the user to validate if it's a working/valid S3 setup.

I understand the docs warn about it, to some extent:

```
Important
OpenEverest does not validate file paths or verify the existence of files in the specified storage buckets. Make sure that the backup directory path is correct and accessible.
To verify that the specified path exists, you can use the AWS CLI:
aws s3 ls s3://<bucket-name>/<path-to-backup> --region <region>
```

But this should definitely be improved. Can we check credentials at least? Even better, check creds + bucket exists. Why not check path too?

If nothing else, at least include the warning in the UI.

### Additional Context

_No response_

### LINKED ISSUE #2245: IsBase64Encoded heuristic breaks inline S3 credentials
### What is the issue?

When a user supplies inline S3 credentials on a `DataImportJob` (or
`DatabaseCluster.spec.dataSource.dataImport`) — i.e. uses the convenience
fields `source.s3.accessKeyId` / `source.s3.secretAccessKey` — the mutating
admission webhook routes the values into a Kubernetes `Secret` based on a
naive heuristic that conflates "looks like base64" with "is base64-encoded".

This causes two distinct problems:

1. **False rejection.** Any pair of values where one happens to look like
   valid base64 and the other does not is rejected with:

   > `failed to create or update S3 credentials secret: both accessKeyID and secretAccessKey must be either base64 encoded or not`

   This is extremely common because typical S3-compatible secret access keys
   are 32 or 40 random alphanumerics (length divisible by 4, all chars in the
   base64 alphabet => flagged as base64 even though they were typed as plain
   text).

2. **Silent credential corruption.** When *both* values happen to look like
   base64, the webhook stores them into `secret.Data` raw rather than
   base64-decoding them. The downstream consumer then reads the *base64
   string itself* as the access key/secret. Authentication fails at runtime
   against S3 instead of at admission time.

The intended behavior is presumably "users may pass either plain values or
pre-base64-encoded values"; the heuristic is the wrong way to detect that.

### Steps to Reproduce

Use the "Import" functionality, and add details for importing from S3-compatible storage in "Provide S3 details". See 1 and 2 above to see which keys trigger these issues.

Where the bug lives:

```go
// utils/validators.go
// IsBase64Encoded checks if the given string is a valid base64 encoded string.
func IsBase64Encoded(s string) bool {
  _, err := base64.StdEncoding.DecodeString(s)
  return len(s)%4 == 0 && len(s) > 0 && err == nil
}
```

This returns `true` for *any* string that is non-empty, has length divisible
by 4, and uses only characters in the standard base64 alphabet
(`A-Z`, `a-z`, `0-9`, `+`, `/`, `=`). It does not distinguish "user typed
base64" from "user typed something that happens to look like base64".

### Relevant Logs / Output

_No response_

### OpenEverest Version

v1.15.2

### Installation Method

Helm

### Kubernetes Distribution

GKE

### Environment Details

_No response_

### Possible Solution

Two possible solutions follow. 

- **Drop the heuristic entirely.** Treat the inline fields as plain values
  always; always populate `StringData`. Document that values must be
  passed in plain form.
- **Make the encoding explicit.** Add a sibling field such as
  `credentialsEncoding: plain | base64` (default `plain`), or rename the
  base64 path to a separate field (`accessKeyIdBase64` /
  `secretAccessKeyBase64`), and add a selector in the UI.

### Additional Context

_No response_

PutlaTheophila:
## Summary

The DataImportJob mutating webhook used `utils.IsBase64Encoded` to decide whether to store inline `AccessKeyID` / `SecretAccessKey` into the credentials Secret as `Data` (raw bytes, treated by the API server as already-base64) or as `StringData` (plain text). The heuristic is

```go
return len(s)%4 == 0 && len(s) > 0 && err == nil
```

`base64.StdEncoding.DecodeString` silently accepts any ASCII string whose length is a multiple of 4 with no internal `=`. Real AWS access keys like

```
AKIAIOSFODNN7EXAMPLEAAAAAAAAAAAA   (length 32)
```

are therefore misclassified as base64 and stored as the decoded byte sequence. The DataImport job then receives a corrupted `AWS_ACCESS_KEY_ID` env var, S3 returns `InvalidAccessKeyId`, and the DB sits in `importing` with no surface-level error.

## Fix

- Drop the heuristic in `mutateS3CredentialsSecret`. Always write user-supplied credentials to `secret.StringData` — the API server encodes them on the write path, so plain strings round-trip unchanged.
- `utils.IsBase64Encoded` is retained because `internal/webhook/enginefeatures.everest/v1alpha1/splithorizondnsconfig_webhook.go` still calls it for CA cert/key validation. Changing that webhook's contract is out of scope here.

## Tests

- `TestDataImportJobDefaulter` now uses 32-char plain-alphanumeric keys (the regression class) and asserts the values land in `StringData` verbatim.
- `TestDatabaseClusterDefaulter` updated for the same assertion — it exercises the same `handleS3CredentialsSecret` path through the `DatabaseCluster` defaulter.
- `TestIsBase64Encoded` removed (the function is no longer called from this package).

## Related

Companion PR in the API server adds a synchronous read-only S3 probe at `DatabaseCluster` admission so the broader "credentials silently fail" class of bug is caught at create time: **openeverest/openeverest#2279** (Fixes openeverest/openeverest#2229).

Together the two PRs make the inline-credentials path correct end-to-end: the operator stores them as plain strings, and the API server validates them against S3 before the resource is admitted.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./internal/webhook/everest/v1alpha1/... ./utils/... ./internal/webhook/enginefeatures.everest/v1alpha1/...` clean
- [x] `go test ./internal/webhook/everest/v1alpha1/... ./utils/... ./internal/webhook/enginefeatures.everest/v1alpha1/...` pass
- [x] DCO sign-off

Fixes openeverest/openeverest#2245.

### DIFF
--- a/internal/webhook/everest/v1alpha1/dataimportjob_defaulter_test.go
+++ b/internal/webhook/everest/v1alpha1/dataimportjob_defaulter_test.go
@@ -78,8 +78,6 @@ func TestDataImportJobDefaulter(t *testing.T) {
 	assert.Empty(t, dij.Spec.DataImportJobTemplate.Source.S3.AccessKeyID)
 	assert.Empty(t, dij.Spec.DataImportJobTemplate.Source.S3.SecretAccessKey)
 
-	// Credentials land in StringData verbatim. Writing to Data with these
-	// values would have stored a corrupted (base64-decoded) byte sequence.
 	secret := &corev1.Secret{}
 	err = client.Get(t.Context(), types.NamespacedName{Namespace: ns, Name: secretName}, secret)
 	require.NoError(t, err)
--- a/internal/webhook/everest/v1alpha1/dataimportjob_defaulter_test.go
+++ b/internal/webhook/everest/v1alpha1/dataimportjob_defaulter_test.go
@@ -27,26 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	everestv1alpha1 "github.com/percona/everest-operator/api/everest/v1alpha1"
-	"github.com/percona/everest-operator/utils"
 )
-
-func TestIsBase64Encoded(t *testing.T) {
-	t.Parallel()
-	cases := []struct {
-		input    string
-		expected bool
-	}{
-		{"", false},
-		{"Zm9v", true},     // "foo" in base64
-		{"Zm9vYmFy", true}, // "foobar" in base64
-		{"notbase64", false},
-		{"Zm9vYmFyIQ==", true}, // "foobar!" in base64
-		{"Zm9vYmFyIQ", false},  // invalid base64 (wrong padding)
-	}
-	for _, c := range cases {
-		assert.Equal(t, c.expected, utils.IsBase64Encoded(c.input), "input: %q", c.input)
-	}
-}
 
 func TestDataImportJobDefaulter(t *testing.T) {
 	t.Parallel()
@@ -57,8 +38,11 @@ func TestDataImportJobDefaulter(t *testing.T) {
 	const (
 		ns         = "test-ns"
 		secretName = "s3-creds"
-		accessKey  = "ZmFrZUFjY2Vzc0tleQ==" // base64 for "fakeAccessKey"
-		secretKey  = "ZmFrZVNlY3JldEtleQ==" //nolint:gosec // base64 for "fakeSecretKey"
+		// 32-char plain-alphanumeric strings: previously misclassified as
+		// base64 by the old IsBase64Encoded heuristic (length % 4 == 0).
+		// Regression case for openeverest/openeverest#2245.
+		accessKey = "AKIAIOSFODNN7EXAMPLEAAAAAAAAAAAA"
+		secretKey = "wJalrXUtnFEMIKAAAAAAAAAAAAAAAAAA" //nolint:gosec
 	)
 
 	dij := &everestv1alpha1.DataImportJob{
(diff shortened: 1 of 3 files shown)


internal/webhook/everest/v1alpha1/dataimportjob_defaulter_test.go
```diff
@@ -90,14 +74,15 @@ func TestDataImportJobDefaulter(t *testing.T) {
...
 	require.NoError(t, err)
 
-	// Check that the credentials are removed from the spec
+	// Credentials are removed from the spec.
 	assert.Empty(t, dij.Spec.DataImportJobTemplate.Source.S3.AccessKeyID)
 	assert.Empty(t, dij.Spec.DataImportJobTemplate.Source.S3.SecretAccessKey)
 
-	// Check that the secret was created and contains the expected data
+	// Credentials land in StringData verbatim. Writing to Data with these
+	// values would have stored a corrupted (base64-decoded) byte sequence.
```
chilagrow: Let's remove this comment, for the readers of the code this comment sounds confusing without context.

chilagrow (commented):
Thanks for your contribution! One tiny comment.

PutlaTheophila:
Thanks for the review @chilagrow! Removed the confusing `StringData`/`Data` comment in a646b73.

chilagrow (commented):
Could you sign-off your commit please? CI is failing

PutlaTheophila:
Thanks @chilagrow — both addressed:

- Comment removed (`6edb5f9`).
- Sign-off added; DCO is green.

chilagrow (commented):
Thanks for your contribution! I run e2e test for data importer a few times and they are failing, could you look at it please https://github.com/openeverest/openeverest-operator/actions/runs/26921958578/job/79426452666?pr=955?

Note, we run these tests manually upon change of the relevant code, so it might have got outdated 🙏

internal/webhook/everest/v1alpha1/dataimportjob_defaulter_test.go
sna0978117341: Hi

PR #959, opened 2026-06-18
Fix pg import race

### LINKED ISSUE #2347: PG backup import marked Succeeded but restores no data
### What is the issue?

A PostgreSQL data import reports `DataImportJob: Succeeded` but the destination database is **empty** when the destination `DatabaseCluster` has `engine.replicas: 2`. It triggers on slow/networked storage; `engine.replicas: 1` works.

- **Expected:** the import populates the destination from the backup, and is marked `Succeeded` only when data is actually restored.
- **Actual:** the destination comes up healthy on a freshly-bootstrapped **empty** volume (its `system_identifier` differs from the source backup), the import is marked `Succeeded`, and `spec.dataSource` is then dropped — so the empty state is **permanent** (no retry).

### Steps to Reproduce

1. Install OpenEverest 1.15.2 (Helm) on a cluster backed by slow/networked storage.
2. Create a source PostgreSQL `DatabaseCluster`, seed a table, and take a backup to S3-compatible storage.
3. Create a destination `DatabaseCluster` with `engine.replicas: 2` and a `dataSource.dataImport` pointing at that backup.
4. Result: `DataImportJob` → `Succeeded`, cluster → `ready`, but the DB is empty (`\dt` shows no tables; `system_identifier` ≠ source).
5. Differentiator: the same import with `engine.replicas: 1` restores correctly.

The trigger is a race: the importer deletes and re-creates the PG cluster to force a bootstrap restore, but only waits for the cluster CR to disappear — not for the old empty bootstrap pods to drain. On slow storage those pods are still serving at re-create, so the cluster comes up on the old empty data. On fast local storage they drain in time and the bug does not occur naturally; it can be forced by making the old bootstrap pods outlive the re-create (e.g. a `preStop` sleep on the bootstrap instances).

If needed, I used the following to simulate slowness:
```
# prestop.sh <cluster-name>   (release with: touch ~/STOP-PRESTOP-<cluster>)
C=$1; FIN_NS=everest
PATCH='{"spec":{"template":{"spec":{"terminationGracePeriodSeconds":120,"containers":[{"name":"database","lifecycle":{"preStop":{"exec":{"command":["bash","-c","sleep 110"]}}}}]}}}}'
while [ ! -f ~/STOP-PRESTOP-$C ]; do
  for s in $(kubectl get sts -n $FIN_NS -l postgres-operator.crunchydata.com/cluster=$C,postgres-operator.crunchydata.com/instance-set=instance1 -o name 2>/dev/null); do
    has=$(kubectl get $s -n $FIN_NS -o jsonpath='{.spec.template.spec.containers[?(@.name=="database")].lifecycle.preStop}' 2>/dev/null)
    [ -z "$has" ] && kubectl patch $s -n $FIN_NS --type=strategic -p "$PATCH" >/dev/null 2>&1
  done
  sleep 0.5
done
```
Run the prestop script before calling the API to import the backup:
```
kubectl create secret generic minio-dataimport-s3-rep7-2 -n everest \
  --from-literal=AWS_ACCESS_KEY_ID=REPLACE-WITH-AWS-ACCESS-KEY \
  --from-literal=AWS_SECRET_ACCESS_KEY=REPLACE-WITH-AWS-SECRET-KEY

kubectl apply -f - <<'EOF'
apiVersion: everest.percona.com/v1alpha1
kind: DatabaseCluster
metadata: { name: restored-rep7-2, namespace: everest }
spec:
  engine: { type: postgresql, version: "18.3", replicas: 2, resources: { cpu: "1", memory: 2Gi }, storage: { size: 2Gi, class: local-path } }
  proxy:  { type: pgbouncer, replicas: 2 }
  dataSource:
    dataImport:
      dataImporterName: everest-percona-pg-operator
      source:
        path: /postgresql-src/1511da0f-cf05-4b96-94ba-0a5d736cb32e/backup/db/20260609-001309F
        s3:
          bucket: everest-backups
          credentialsSecretName: minio-dataimport-s3-rep7-2
          endpointURL: https://minio-service.default.svc.cluster.local
          forcePathStyle: true
          region: us-east-1
          verifyTLS: false
EOF
```

### Relevant Logs / Output

```text
# DataImportJob succeeded and cluster is ready...
$ kubectl get dataimportjob data-import-restored          # STATE = Succeeded
$ kubectl get db restored                                 # READY 4/4, STATUS ready

# ...but the dataSource was stripped (one-shot, no retry):
$ kubectl get pg restored -o jsonpath='{.spec.dataSource}'   # (empty)

# ...and the primary is empty with a system_identifier that does NOT match the source backup:
$ kubectl exec restored-<primary> -c database -- psql -U postgres -c '\dt'
Did not find any relations.
$ kubectl exec restored-<primary> -c database -- psql -U postgres -At \
    -c "SELECT system_identifier FROM pg_control_system()"   # != source backup's sysid
```

### OpenEverest Version

v1.15.2

### Installation Method

Helm

### Kubernetes Distribution

Other (please specify below)

### Environment Details

| Component | Version |
|-----------|---------|
| OpenEverest server | 1.15.2 |
| everest-operator / DataImporter image | 1.15.2 |
| percona-postgresql-operator (PGO) | 2.9.0 |
| PostgreSQL engine | 18.3 |

- **Kubernetes distribution:** N/A — the defect is distribution-agnostic. Reproduced on a local k3s-in-Docker (k3d); occurs on any distribution where pod/volume drain is slow.
- **Storage:** the environmental trigger is **slow/networked storage**. Fast local storage hides the bug (old instances drain before the re-create).
- **Backups:** S3-compatible object storage.

### Possible Solution

In the importer's `restorePGCluster` delete→re-create sequence, wait for the **old instance Pods/PVCs to fully drain** before re-creating — not just for the `PerconaPGCluster` CR to become `NotFound`.

Additionally, the import should only be considered successful when the restore actually populated data, rather than marking `Succeeded` purely because the import Job exited 0 and the cluster reached `AppStateReady` (a healthy **empty** cluster satisfies that state).

### Additional Context

**Root cause.** For PG, the importer must set `spec.dataSource` at create time, so it deletes and re-creates the already-bootstrapped (empty) cluster. The re-create polls only until the `PerconaPGCluster` CR is `NotFound`, then re-creates immediately — while the old empty bootstrap pods may still be Running. PGO then comes up on that surviving empty instance (no fresh restore). The cluster reaches `AppStateReady` (which only checks ready/size counts, satisfied by a healthy empty cluster), the `DataImportJob` is marked `Succeeded` on Job exit 0, and `spec.dataSource` is subsequently dropped. The bootstrap restore is a one-shot create-time operation, so it is never retried, and left permanently empty. `engine.replicas: 2` produces two pods to drain, which on slow storage reliably outlive the re-create; `replicas: 1` drains in time, so the restore runs into a fresh volume.

**PGO is not at fault:** a fresh-create `PerconaPGCluster` with `dataSource` at `replicas: 2` restores correctly. The bug is the importer's delete→re-create sequencing.

The defect is in the everest-operator PG data importer (`internal/data-importer/cmd/pg/import.go`, `restorePGCluster`) and the DataImportJob controller (`dataimportjob_controller.go`). The OpenEverest server/API/UI is not involved.

oksana-grishchenko:
**Fix pg import race**
---
**Problem:**
fixes https://github.com/openeverest/openeverest/issues/2347 

During import, the pg CR gets recreated with DataSource. The CR itself gets recreated quickly however the related resources (pods, pvcs) could take awhile to be deleted - on slow storages it could happen that CR reuses the old empty resources and comes up ready.  

**Solution:**
Ensure the old resources are gone before recreating pg CR

Related PRs:
- https://github.com/openeverest/helm-charts/pull/59

**CHECKLIST**
---
**Tests**


recharte (approved):
Thanks Oksana

PR #961, opened 2026-06-26
Allow separate CPU/Memory requests and limits for database engines and proxies

### LINKED ISSUE #2360: Allow to set requests and limits separately
### Is your feature request related to a problem?

Right now when the user is creating a database in v1, there are fields to set CPU and RAM for database pods. Setting these fields is translated into `requests` and `limits`. These are set equal to ensure guaranteed QoS as per the best practices for stateful workloads.
There are still cases where users might need to set requests a bit lower and users are fine with these Pods to be evicted or killed. Example, non-production workloads. 

### Describe the solution you'd like

It would be cool if API/UI allow users to set requests and limits separately. Maybe default to single value, but enable some advanced settings where these values can be set separately. 

### How would users interact with this feature?

UI/API, later on CLI as well.

### Alternative solutions considered

_No response_

### Additional Context

It is interesting that even v2 unblocks this, it might be disregarded by Provider builders. It is okay, but would be great to have it in there as well to ensure proper flexibility.

recharte:
## Summary

Enable users to specify separate resource **requests** and **limits** for database engines and proxies, instead of being forced to set them to the same value. Maintains full backward compatibility: existing clusters using the flat `cpu`/`memory` fields continue to work identically.

## Changes

- **API types** (`api/everest/v1alpha1/databasecluster_types.go`):
  - Add `ResourceSpec` leaf type for CPU/Memory pairs
  - Extend `Resources` struct with optional `Limits` and `Requests` sub-objects
  - Mark flat `CPU`/`Memory` fields as deprecated (now `+optional`)
  - Add helpers: `EffectiveLimits()`, `EffectiveRequests()`, `ToResourceRequirements(mirror bool)`
  - Update `Engine.Size()` to use effective limits memory for sizing

- **Reconcilers**:
  - PXC (engine, HAProxy, ProxySQL): route resources through `ToResourceRequirements(mirror=true)` to preserve legacy behavior (requests=limits)
  - PSMDB (replset, mongos): use `ToResourceRequirements(mirror=false)` for limits-only defaults
  - PostgreSQL (engine, pgbouncer): use `ToResourceRequirements(mirror=false)` for limits-only defaults

## Backward Compatibility

**Fully backward compatible**: 
- Existing CRs using only `cpu`/`memory` produce identical upstream specs
- PXC continues to set requests=limits when only legacy fields present
- PSMDB/PostgreSQL continue to set limits-only when only legacy fields present
- No migration or manual intervention required

ref: https://github.com/openeverest/openeverest/issues/2360


internal/controller/everest/providers/pxc/applier.go
```diff
@@ -304,13 +304,14 @@ func (p *applier) Engine() error {
...
 
-	if !p.DB.Spec.Engine.Resources.CPU.IsZero() {
-		pxc.Spec.PXC.PodSpec.Resources.Limits[corev1.ResourceCPU] = p.DB.Spec.Engine.Resources.CPU
-		pxc.Spec.PXC.PodSpec.Resources.Requests[corev1.ResourceCPU] = p.DB.Spec.Engine.Resources.CPU
-	}
-	if !p.DB.Spec.Engine.Resources.Memory.IsZero() {
-		pxc.Spec.PXC.PodSpec.Resources.Limits[corev1.ResourceMemory] = p.DB.Spec.Engine.Resources.Memory
-		pxc.Spec.PXC.PodSpec.Resources.Requests[corev1.ResourceMemory] = p.DB.Spec.Engine.Resources.Memory
+	{
+		engineResources := p.DB.Spec.Engine.Resources.ToResourceRequirements(true)
```
chilagrow: Could we double check the below logic is what we want? 🤔

* input `{limits: {cpu:3, memory: 4GB}, requests: {cpu:5}}`
* currently sets `{limits: {cpu:3, memory: 4GB}, requests: {cpu:5, memory: 4GB}}`
* we don't want this right? `{limits: {cpu:3, memory: 4GB}, requests {cpu:5}}`

chilagrow (approved):

internal/controller/everest/providers/pxc/applier.go
recharte: We definitely don't want that, we shouldn't even allow the requests to be larger than the limits.

chilagrow: Oops my example came out very wrong 😓

What I really meant was, if requests does not specify either CPU or Memory, we do want to copy just the missing individual value such as `limits.memory` to `requests.memory` right?

recharte: After looking at this more carefully I decided to do a big refactoring that not only handles the case you mentioned but also removes duplicate code and methods that shouldn't exist at the API layer.

chilagrow (approved):
Thank you!

